PR #18, opened 2019-06-30
Added dutch translations

mjarkk:
Title tells enough.


pkg/i18n/dutch.go
```diff
@@ -1,5 +1,90 @@
...
+		EditConfig:         "verander de lazydocker configuratie",
+		Cancel:             "annuleren",
+		Remove:             "verwijder",
+		ForceRemove:        "geforceerd verwijderen",
+		RemoveWithVolumes:  "verwijder met volumes",
+		RemoveService:      "verwijder containers",
+		Stop:               "stop",
+		Restart:            "herstart",
+		Rebuild:            "herbouw",
+		Recreate:           "recreate",
```
mjarkk: I need a bit of help for *recreate*.  
I can't think of a good word for this aside from *herbouwen* what we already used for rebuild.  
Google translate suggests *herscheppen* but that sounds a bit to religious to me

YannickMeeus: How about "hercreëer"? It's as close to the English, and while it's been a while since I considered Dutch to be my main language it still *feels* okay to me.

mjarkk: Thanks, i'll use that

.vscode/settings.json
```diff
@@ -0,0 +1,7 @@
+{
```
jesseduffield: if you rebase this onto master, I'm guessing it will remove this file :)

mjarkk:
Shall i do a *squash and merge*, I don't want to crap to much around with the master branch.

jesseduffield:
that sounds good to me :)

PR #19, opened 2019-06-30
Tweak installation instructions when using Go

davestephens:
The installation instructions when using Go to grab the code were missing the key `go install` step. I've added this to the readme, as well as pointed out that adding the go workspace bin path to your profile may be desirable.

Cool project, thanks for your work on it! :+1:

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,10 @@ required go version: 1.12
 
 ```sh
 go get github.com/jesseduffield/lazydocker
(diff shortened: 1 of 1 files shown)


mjarkk:
I don't see why the `go install` is needed.  
when a user has a correct `GOPATH` go will do a `go install` automatically when using `go get`

davestephens:
You are right! I'll close this...

PR #21, opened 2019-06-30
German translation

waterfl0w:
This PR provides a German translation for the key bindings documentation, and the software itself. If there are more files that can be translated at this point in time, please feel free to point me at them and I will try to accommodate some more time to translate them.


mjarkk:
Amazing!  
If you want to help further jesseduffield also has a project called [lazygit](https://github.com/jesseduffield/lazygit) that also doesn't have a German translation yet

jesseduffield:
Thankyou so much this is awesome!

Just one more thing required I think: you'll need to chuck in something like this in NewTranslationSet in `pkg/i18n/i18n.go`:

```
	if strings.HasPrefix(userLang, "nl") {
		mergo.Merge(&set, polishSet(), mergo.WithOverride)
	}
```

Issue #13, opened 2019-06-30
go get fails currently


EinPinsel:
While trying to install lazydocker, it fails with the following error: "undefined: strings.ReplaceAll" in 
https://github.com/jesseduffield/lazydocker/blob/51d9b14ab5549a76fc159bbe0ac3e5d387a8ee00/pkg/commands/image.go#L71

mjarkk:
What is your go version?

jesseduffield:
@EinPinsel apologies, it turns out that function is only available in go1.12 onwards. I added to the readme that you need that version. I'm not really averse to using a different function than strings.ReplaceAll if it's enough of a hassle for people who are on old go versions. What are your thoughts?

EinPinsel:
You got me there - it was `go version go1.10.4 linux/amd64`. With 1.12 it works. I'm still finding my way around go - and that you can't just get it from the default repo is a bit of missing comfort.

jesseduffield:
@EinPinsel fair enough. I think I'll close this issue and if we get another person who gets stumped on this I'll go and replace the function with something else :)

PR #32, opened 2019-07-01
fix strings.ReplaceAll() does't exist error while go version bellow 1.12

jonyhy96:
### reason
I also got strings.ReplaceAll does't exist error while go get so i want fix it

### version
- go 1.11.9 linux/amd64
- docker 18.06.1-ce

### issue
#13 

### solution
use strings.Replace() instead


jesseduffield:
thanks @jonyhy96 !

Issue #29, opened 2019-07-01
Not able to install in windows


andybala:
Hi all,

tried to install on Windows with latest goland install. errors attached, not sure what's going wrong

# go get github.com/jesseduffield/lazydocker
# github.com/jesseduffield/lazydocker/pkg/commands
C:\Users\AHB\go\src\github.com\jesseduffield\lazydocker\pkg\commands\os.go:377:46: cmd.SysProcAttr.Setpgid undefined (type *syscall.SysProcAttr has no field or method Setpgid)
C:\Users\AHB\go\src\github.com\jesseduffield\lazydocker\pkg\commands\os.go:379:10: undefined: syscall.Kill
C:\Users\AHB\go\src\github.com\jesseduffield\lazydocker\pkg\commands\os.go:388:3: unknown field 'Setpgid' in struct literal of type syscall.SysProcAttr

-- andy

jesseduffield:
Would it suffice to run this in WSL? There are some tricky things I would need to fix up to get this to run in windows itself

mohamedahmed01:
@jesseduffield 
It seems this code would allow you to bootstrap windows support but i dont know enough go lang enough to implmement it , would you check it out.
https://blog.csdn.net/fyxichen/article/details/51857864

jesseduffield:
@mohamedahmed01 the issue with that code is that I'm using SetPgid so that I can group child processes of a process I run. An example is `docker-compose logs` which itself spawns one process for each service, all of which print to stdout. I give them all the same process group id so that I can kill the group when I want to show the output of some other command. By just returning nil I'd be losing that functionality.

from that link:
```
func SetPgid(pid, pgid int) error {
	return nil
}
```

I believe this is the only real think preventing windows from working, so a proper solution may not be too hard to achieve. Something along the lines of looking at the process table to find all descendants of the process we want to kill could suffice.

jenyayel:
After installing it in WSL and starting, it crashes:
```
[...]

Issue #28, opened 2019-07-01
Provide a docker image option


sandys:
A similar tool Dry - https://github.com/moncho/dry - provides the capability to be run using docker itself

```docker run --rm -it -v /var/run/docker.sock:/var/run/docker.sock -e DOCKER_HOST=$DOCKER_HOST moncho/dry```

can you please also provide a docker container for this ?

**Comment by mjarkk** [comment]
Docui also has a docker run command: `docker run --rm -itv /var/run/docker.sock:/var/run/docker.sock skanehira/docui`

At the moment i think this is not possible due to missing feather in gocui, (For details see https://github.com/awesome-gocui/gocui/pull/18)  

If that fix is merged into jesses's gocui fork we could make a docker image for lazydocker too.

**Comment by dawidd6** [comment]
@jesseduffield I think @mjarkk is right. We should apply this patch on gocui, lazygit can benefit from it too and there would be no need to make some workarounds in Dockerfile or twice in two codebases.

**Comment by dawidd6** [comment]
https://github.com/jesseduffield/lazydocker/pull/36

**Comment by qdm12** [comment]
I'm almost done fixing one or two last issues for the pull request to be merged, it should be done on my side by tonight.

**Comment by jesseduffield** [comment]
I only just read this thread then. So to be clear, if we merge that gocui PR (https://github.com/awesome-gocui/gocui/pull/18/files) into my gocui fork, we won't need to add any window-size checking code in https://github.com/jesseduffield/lazydocker/pull/36? Is that correct @dawidd6 @qdm12 ?

If so I'm happy to go ahead and do that now

**Comment by qdm12** [comment]
It could be, for now we have the following code to check on the terminal size:

```sh
func waitForTerminalSpace() error {
	// before we do anything, we need to check that we have some window space available
	width, height, err := terminal.GetSize(int(os.Stdin.Fd()))
	if err != nil {
		return err
	}
	if width > 0 && height > 0 {
		return nil
	}
	winch := make(chan os.Signal)
	signal.Notify(winch, syscall.SIGWINCH)
	select {
	case <-winch:
		return nil
	case <-time.After(time.Second):
		return fmt.Errorf("there is no available terminal space")
	}
}
```

And it works pretty well I think. Should I commit it now or wait for you to implement changes to your gocui fork?

jesseduffield:
I think the gocui fork is the right approach. I've got a PR up here https://github.com/jesseduffield/lazydocker/pull/103

[...]

PR #36, opened 2019-07-01
Reworked Dockerfile for a more production oriented usage

qdm12:
- Added .dockerignore to speed up build context and avoid rebuilding when unecessary
- Specified Alpine and Go versions as build arguments
- Specified Go target CPU architecture as build arguments to be able to build for ARM devices
- Specified base images as build arguments to be able to build for ARM devices
- Trimmed down size of final image using `-a -installsuffix cgo -ldflags="-s -w"` go build flags and by copying the statically built binary only to the final image
- Added clear build and run instructions for the Docker container

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,6 +24,5 @@ LABEL org.label-schema.schema-version="1.0.0-rc1" \
     org.label-schema.vcs-description="The lazier way to manage everything docker" \
     org.label-schema.docker.cmd="docker run -it -v /var/run/docker.sock:/var/run/docker.sock lazydocker" \
     org.label-schema.version=${VERSION}
-RUN apk --update add -q --progress --no-cache -U git xdg-utils
-ENTRYPOINT [ "lazydocker" ]
+ENTRYPOINT [ "/bin/sh" ]
 COPY --from=builder /go/src/github.com/jesseduffield/lazydocker/lazydocker /usr/bin/lazydocker
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,4 +27,5 @@ LABEL org.label-schema.schema-version="1.0.0-rc1" \
     org.label-schema.docker.cmd="docker run -it -v /var/run/docker.sock:/var/run/docker.sock lazydocker" \
     org.label-schema.version=${VERSION}
 ENTRYPOINT [ "/lazydocker" ]
+VOLUME [ "/.config/jesseduffield/lazydocker" ]
 COPY --from=builder /go/src/github.com/jesseduffield/lazydocker/lazydocker /lazydocker
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,11 +6,24 @@ ARG GO_VERSION=1.12.6
 FROM ${BASE_IMAGE_BUILDER}:${GO_VERSION}-alpine${ALPINE_VERSION} AS builder
 ARG GOARCH=amd64
 ARG GOARM
+ARG VERSION=0.2.4
+RUN apk --update add -q --progress git
 WORKDIR /go/src/github.com/jesseduffield/lazydocker/
 COPY ./ .
-RUN CGO_ENABLED=0 GOOS=linux GOARCH=${GOARCH} GOARM=${GOARM} go build -a -installsuffix cgo -ldflags="-s -w"
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=${GOARCH} GOARM=${GOARM} go build -a -installsuffix cgo -ldflags="-s -w \
+    -X main.commit=$(git rev-parse HEAD) \
+    -X main.version=${VERSION} \
+    -X main.buildSource=Docker"
 
 FROM ${BASE_IMAGE}:${ALPINE_VERSION}
+ARG VERSION=0.2.4
+LABEL org.label-schema.schema-version="1.0.0-rc1" \
+    maintainer="jessedduffield@gmail.com" \
+    org.label-schema.vcs-ref=$VCS_REF \
+    org.label-schema.url="https://github.com/jesseduffield/lazydocker" \
+    org.label-schema.vcs-description="The lazier way to manage everything docker" \
+    org.label-schema.docker.cmd="docker run -it -v /var/run/docker.sock:/var/run/docker.sock lazydocker" \
+    org.label-schema.version=${VERSION}
 RUN apk --update add -q --progress --no-cache -U git xdg-utils
 ENTRYPOINT [ "lazydocker" ]
 COPY --from=builder /go/src/github.com/jesseduffield/lazydocker/lazydocker /usr/bin/lazydocker
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,7 +9,7 @@ ARG VERSION
 ARG VCS_REF
 WORKDIR /go/src/github.com/jesseduffield/lazydocker/
 COPY ./ .
-RUN CGO_ENABLED=0 GOOS=linux GOARCH=${GOARCH} GOARM=${GOARM} go build -a -installsuffix cgo -ldflags="-s -w \
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=${GOARCH} GOARM=${GOARM} go build -a -ldflags="-s -w \
     -X main.commit=${VCS_REF} \
     -X main.version=${VERSION} \
     -X main.buildSource=Docker"
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,8 +6,8 @@ ARG GO_VERSION=1.12.6
 FROM ${BASE_IMAGE_BUILDER}:${GO_VERSION}-alpine${ALPINE_VERSION} AS builder
 ARG GOARCH=amd64
 ARG GOARM
-ARG VERSION=0.2.4
-ARG VCS_REF=
+ARG VERSION
+ARG VCS_REF
 WORKDIR /go/src/github.com/jesseduffield/lazydocker/
 COPY ./ .
 RUN CGO_ENABLED=0 GOOS=linux GOARCH=${GOARCH} GOARM=${GOARM} go build -a -installsuffix cgo -ldflags="-s -w \
@@ -16,8 +16,8 @@ RUN CGO_ENABLED=0 GOOS=linux GOARCH=${GOARCH} GOARM=${GOARM} go build -a -instal
     -X main.buildSource=Docker"
 
 FROM ${BASE_IMAGE}:${ALPINE_VERSION}
-ARG VERSION=0.2.4
-ARG VCS_REF=
+ARG VERSION
+ARG VCS_REF
 LABEL org.label-schema.schema-version="1.0.0-rc1" \
     maintainer="jessedduffield@gmail.com" \
     org.label-schema.vcs-ref=$VCS_REF \
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,9 +7,11 @@ ARG GOARCH=amd64
 ARG GOARM
 ARG VERSION
 ARG VCS_REF
-WORKDIR /go/src/github.com/jesseduffield/lazydocker/
+WORKDIR /tmp/gobuild
 COPY ./ .
-RUN CGO_ENABLED=0 GOOS=linux GOARCH=${GOARCH} GOARM=${GOARM} go build -a -ldflags="-s -w \
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=${GOARCH} GOARM=${GOARM} \
+    go build -a -mod=vendor \
+    -ldflags="-s -w \
     -X main.commit=${VCS_REF} \
     -X main.version=${VERSION} \
     -X main.buildSource=Docker"
@@ -43,5 +45,5 @@ LABEL org.label-schema.schema-version="1.0.0-rc1" \
 ENTRYPOINT [ "/bin/lazydocker" ]
 VOLUME [ "/.config/jesseduffield/lazydocker" ]
 ENV PATH=/bin
-COPY --from=builder /go/src/github.com/jesseduffield/lazydocker/lazydocker /bin/lazydocker
 COPY --from=docker-builder /go/src/github.com/docker/cli/build/docker /bin/docker
+COPY --from=builder /tmp/gobuild/lazydocker /bin/lazydocker
--- a/Dockerfile
+++ b/Dockerfile
@@ -45,6 +45,5 @@ LABEL \
     org.opencontainers.image.title="lazydocker" \
     org.opencontainers.image.description="The lazier way to manage everything docker"
 ENTRYPOINT [ "/bin/lazydocker" ]
-ENV PATH=/bin
 COPY --from=docker-builder /go/src/github.com/docker/cli/build/docker /bin/docker
 COPY --from=builder /tmp/gobuild/lazydocker /bin/lazydocker
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,5 +25,6 @@ LABEL org.label-schema.schema-version="1.0.0-rc1" \
     org.label-schema.vcs-description="The lazier way to manage everything docker" \
     org.label-schema.docker.cmd="docker run -it -v /var/run/docker.sock:/var/run/docker.sock lazydocker" \
     org.label-schema.version=${VERSION}
-ENTRYPOINT [ "/bin/sh" ]
+RUN printf "#!/bin/sh\n\nresize > /dev/null\nlazydocker \$@\n" > /usr/bin/run && chmod 500 /usr/bin/run
+ENTRYPOINT [ "/usr/bin/run" ]
 COPY --from=builder /go/src/github.com/jesseduffield/lazydocker/lazydocker /usr/bin/lazydocker
--- a/Dockerfile
+++ b/Dockerfile
@@ -40,7 +40,8 @@ LABEL org.label-schema.schema-version="1.0.0-rc1" \
     org.label-schema.vcs-description="The lazier way to manage everything docker" \
     org.label-schema.docker.cmd="docker run -it -v /var/run/docker.sock:/var/run/docker.sock lazydocker" \
     org.label-schema.version=${VERSION}
-ENTRYPOINT [ "/lazydocker" ]
+ENTRYPOINT [ "/bin/lazydocker" ]
 VOLUME [ "/.config/jesseduffield/lazydocker" ]
-ENV PATH=/
-COPY --from=docker-builder /go/src/github.com/docker/cli/build/docker /docker
+ENV PATH=/bin
+COPY --from=builder /go/src/github.com/jesseduffield/lazydocker/lazydocker /bin/lazydocker
+COPY --from=docker-builder /go/src/github.com/docker/cli/build/docker /bin/docker
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,15 +1,16 @@
-# run with:
-# docker build -t lazydocker .
-# docker run -it lazydocker:latest /bin/sh -l
+ARG BASE_IMAGE_BUILDER=golang
+ARG BASE_IMAGE=alpine
+ARG ALPINE_VERSION=3.10
+ARG GO_VERSION=1.12.6
 
-FROM golang:alpine
+FROM ${BASE_IMAGE_BUILDER}:${GO_VERSION}-alpine${ALPINE_VERSION} AS builder
+ARG GOARCH=amd64
+ARG GOARM
 WORKDIR /go/src/github.com/jesseduffield/lazydocker/
 COPY ./ .
-RUN CGO_ENABLED=0 GOOS=linux go build
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=${GOARCH} GOARM=${GOARM} go build -a -installsuffix cgo -ldflags="-s -w"
 
-FROM alpine:latest
-RUN apk add -U git xdg-utils
-WORKDIR /go/src/github.com/jesseduffield/lazydocker/
-COPY --from=0 /go/src/github.com/jesseduffield/lazydocker /go/src/github.com/jesseduffield/lazydocker
-COPY --from=0 /go/src/github.com/jesseduffield/lazydocker/lazydocker /bin/
-RUN echo "alias gg=lazydocker" >> ~/.profile
+FROM ${BASE_IMAGE}:${ALPINE_VERSION}
+RUN apk --update add -q --progress --no-cache -U git xdg-utils
+ENTRYPOINT [ "lazydocker" ]
+COPY --from=builder /go/src/github.com/jesseduffield/lazydocker/lazydocker /usr/bin/lazydocker
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,16 +7,17 @@ FROM ${BASE_IMAGE_BUILDER}:${GO_VERSION}-alpine${ALPINE_VERSION} AS builder
 ARG GOARCH=amd64
 ARG GOARM
 ARG VERSION=0.2.4
-RUN apk --update add -q --progress git
+ARG VCS_REF=
 WORKDIR /go/src/github.com/jesseduffield/lazydocker/
 COPY ./ .
 RUN CGO_ENABLED=0 GOOS=linux GOARCH=${GOARCH} GOARM=${GOARM} go build -a -installsuffix cgo -ldflags="-s -w \
-    -X main.commit=$(git rev-parse HEAD) \
+    -X main.commit=${VCS_REF} \
     -X main.version=${VERSION} \
     -X main.buildSource=Docker"
 
 FROM ${BASE_IMAGE}:${ALPINE_VERSION}
 ARG VERSION=0.2.4
+ARG VCS_REF=
 LABEL org.label-schema.schema-version="1.0.0-rc1" \
     maintainer="jessedduffield@gmail.com" \
     org.label-schema.vcs-ref=$VCS_REF \
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,4 @@
 ARG BASE_IMAGE_BUILDER=golang
-ARG BASE_IMAGE=alpine
 ARG ALPINE_VERSION=3.10
 ARG GO_VERSION=1.12.6
 
@@ -15,16 +14,17 @@ RUN CGO_ENABLED=0 GOOS=linux GOARCH=${GOARCH} GOARM=${GOARM} go build -a -instal
     -X main.version=${VERSION} \
     -X main.buildSource=Docker"
 
-FROM ${BASE_IMAGE}:${ALPINE_VERSION}
-ARG VERSION
+FROM scratch
+ARG BUILD_DATE
 ARG VCS_REF
+ARG VERSION
 LABEL org.label-schema.schema-version="1.0.0-rc1" \
     maintainer="jessedduffield@gmail.com" \
+    org.label-schema.build-date=$BUILD_DATE \
     org.label-schema.vcs-ref=$VCS_REF \
     org.label-schema.url="https://github.com/jesseduffield/lazydocker" \
     org.label-schema.vcs-description="The lazier way to manage everything docker" \
     org.label-schema.docker.cmd="docker run -it -v /var/run/docker.sock:/var/run/docker.sock lazydocker" \
     org.label-schema.version=${VERSION}
-RUN printf "#!/bin/sh\n\nresize > /dev/null\nlazydocker \$@\n" > /usr/bin/run && chmod 500 /usr/bin/run
-ENTRYPOINT [ "/usr/bin/run" ]
-COPY --from=builder /go/src/github.com/jesseduffield/lazydocker/lazydocker /usr/bin/lazydocker
+ENTRYPOINT [ "/lazydocker" ]
+COPY --from=builder /go/src/github.com/jesseduffield/lazydocker/lazydocker /lazydocker
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,6 +14,20 @@ RUN CGO_ENABLED=0 GOOS=linux GOARCH=${GOARCH} GOARM=${GOARM} go build -a -ldflag
     -X main.version=${VERSION} \
     -X main.buildSource=Docker"
 
+FROM ${BASE_IMAGE_BUILDER}:${GO_VERSION}-alpine${ALPINE_VERSION} AS docker-builder
+ARG GOARCH=amd64
+ARG GOARM
+ARG DOCKER_VERSION=v18.09.7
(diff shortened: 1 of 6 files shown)


qdm12:
Hi there !

Hi ! I just added that so I can run your program without downloading the binary, especially for die hard of running everything in a Docker container... And added ARM build support too. By the way, it seems your program requires Git to function, is this expected? If not, we could manage to run it in a Scratch container as a static binary which would make it very tiny 👍

dawidd6:
Hmm, I think that needing git to function is a bug. The initial codebase was taken from lazygit and maybe something wasn't removed yet.

LGTM. Additionally you may want to add these to `go build`:
```
-ldflags \
		"-X main.commit=$(COMMIT) \
		-X main.date=$(shell date -u +%Y-%m-%d) \
		-X main.version=0.2.4 \
		-X main.buildSource=Docker"
```
I mean, really only version, commit and buildSource would be enough, cause specyfing date could destroy reproducible building.

qdm12:
Done, I've added Docker labels with these as well, it might turn to be useful.
By the way, the error without git is:

```
*errors.errorString invalid dimensions
/go/src/github.com/jesseduffield/lazydocker/vendor/github.com/jesseduffield/gocui/gui.go:164 (0x7d1339)
/go/src/github.com/jesseduffield/lazydocker/pkg/gui/layout.go:84 (0x8010e4)
/go/src/github.com/jesseduffield/lazydocker/vendor/github.com/jesseduffield/gocui/gui.go:387 (0x7d2d90)
/go/src/github.com/jesseduffield/lazydocker/vendor/github.com/jesseduffield/gocui/gui.go:504 (0x7d389e)
/go/src/github.com/jesseduffield/lazydocker/vendor/github.com/jesseduffield/gocui/gui.go:412 (0x7d2fe4)
/go/src/github.com/jesseduffield/lazydocker/pkg/gui/gui.go:309 (0x7f70f3)
/go/src/github.com/jesseduffield/lazydocker/pkg/gui/subprocess.go:20 (0x80725f)
/go/src/github.com/jesseduffield/lazydocker/pkg/app/app.go:54 (0x81a71a)
/go/src/github.com/jesseduffield/lazydocker/pkg/app/app.go:54 (0x81a70d)
/usr/local/go/src/runtime/proc.go:200 (0x42d77c)
/usr/local/go/src/runtime/asm_amd64.s:1337 (0x458541)
```

Also, is `xdg-utils` necessary? It works without for me but I wanted to check with you first.

Thanks !

qdm12:
ok so it seems both `git` and `xdg-utils` are not necessary.

However, running the container directly without first starting a shell `/bin/sh` fails because of the error of gocui above, probably because the screen size is zero or something.

So I will update the instructions, where you need to launch the container and then enter `lazydocker` for the GUI to work out.

qdm12:
It works great on my linux host and my windows host, let me know if I should change anything !

Dockerfile
```diff
@@ -1,15 +1,28 @@
...
-RUN apk add -U git xdg-utils
-WORKDIR /go/src/github.com/jesseduffield/lazydocker/
-COPY --from=0 /go/src/github.com/jesseduffield/lazydocker /go/src/github.com/jesseduffield/lazydocker
-COPY --from=0 /go/src/github.com/jesseduffield/lazydocker/lazydocker /bin/
-RUN echo "alias gg=lazydocker" >> ~/.profile
+FROM ${BASE_IMAGE}:${ALPINE_VERSION}
+ARG VERSION=0.2.4
+LABEL org.label-schema.schema-version="1.0.0-rc1" \
+    maintainer="jessedduffield@gmail.com" \
+    org.label-schema.vcs-ref=$VCS_REF \
```
jesseduffield: should $VCS_REF be defined above?

jesseduffield:
Hey man thanks for making this :) Very impressive.

I left one comment, other than that this LGTM, happy to merge if @dawidd6 is happy with it

Dockerfile
qdm12: Oh yes I'll remove that line, that's for automated Docker builds really... By the way, would you have any idea why gocui fails with the error above? Because it would be nice to have the container run directly lazydocker as its entrypoint :)

jesseduffield: it's based on this:
```
func (g *Gui) SetView(name string, x0, y0, x1, y1 int, overlaps byte) (*View, error) {
	if x0 >= x1 {
		return nil, errors.New("invalid dimensions")
```

Seems like it's an issue with the window being too small. Perhaps the program starts before the terminal is attached. We might need to add some code in saying to wait until the window size is large enough before calling SetView

README.md
```diff
@@ -48,6 +48,57 @@ required go version: 1.12
...
 ```
 
+### Docker
+
+1. <details><summary>Build it...</summary><p>
+
+    - If you have a x86_64 CPU architecture
+
+        ```sh
+        docker build -t lazydocker . https://github.com/jesseduffield/lazydocker.git
```
dawidd6: Umm, `docker build` accepts only one argument: `PATH or URL or -`, so here maybe we should to split this command to two separate ones like:

`docker build -t lazydocker .`
> if building while being in source tree

` docker build -t lazydocker https://github.com/jesseduffield/lazydocker.git`
> if building without source tree

README.md
```diff
@@ -48,6 +48,57 @@ required go version: 1.12
...
+        --build-arg BASE_IMAGE=arm64v8/alpine \
+        --build-arg GOARCH=arm64 \
+        https://github.com/jesseduffield/lazydocker.git
+        ```
+
+    </p></details>
+1. Run it
+
+    ```sh
+    docker run -it -v /var/run/docker.sock:/var/run/docker.sock lazydocker
```
dawidd6: Maybe we can also provide an option to mount config file?

dawidd6:
There seems to be a problem with getting containers logs. It works with normal installation, but not in Docker. @jesseduffield are the logs fetched from command line Docker client?

dawidd6:
```diff
diff --git a/Dockerfile b/Dockerfile
index 351e334..0abcc0f 100644
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,5 +24,6 @@ LABEL org.label-schema.schema-version="1.0.0-rc1" \
     org.label-schema.vcs-description="The lazier way to manage everything docker" \
     org.label-schema.docker.cmd="docker run -it -v /var/run/docker.sock:/var/run/docker.sock lazydocker" \
     org.label-schema.version=${VERSION}
-ENTRYPOINT [ "/bin/sh" ]
+RUN printf "resize > /dev/null\nlazydocker \$@\n" > /bin/run && chmod +x /bin/run
+ENTRYPOINT [ "sh", "/bin/run" ]
 COPY --from=builder /go/src/github.com/jesseduffield/lazydocker/lazydocker /usr/bin/lazydocker
diff --git a/README.md b/README.md
index c651f8f..0af4172 100644
--- a/README.md
+++ b/README.md
@@ -97,8 +97,6 @@ go get github.com/jesseduffield/lazydocker
     docker run -it -v /var/run/docker.sock:/var/run/docker.sock lazydocker
     ```
 
-1. Then enter `lazydocker`
-
 ## Usage
 
 Call `lazydocker` in your terminal. I personally use this a lot so I've made an alias for it like so:
```

@qdm12 you can apply this patch on top of your sources in `reworking-dockerfile` branch if you want to immediately run `lazydocker` instead of first interactively running `sh`.

jesseduffield:
@dawidd6 yes they are fetched vis the command line. Though there are various places in the app that use the docker client package to do things. Perhaps there are issues due to that?

dawidd6:
@jesseduffield heh, so now there are 2 possibilities:
1. Try to rewrite those calls in pure Go code.
2. Install docker package in container.

Obviously the second one is an instant fix, but in the long run, would be better to do the first. Though I don't really know if it's feasible, would need to dig deeper into the code.

Dockerfile
skanehira: I think If you want to run the container that’s need to  check and wait SIGWINCH signal

https://github.com/skanehira/docui/blob/08d7aa1ad7e5f417cffeba9fc2d5a627fda74221/common/common.go#L194

README.md
qdm12: Oh yes typo, my bad ! I guess it's easier for the user to build directly from the git repo, should we just keep it that way? The user probably knows how to git clone and then docker build the repo if he wants to take the long way I think.

README.md
qdm12: Yes good idea, I think it's in the `/.config` directory from what I've seen? Or can we specify its location?

Dockerfile
dawidd6: That's nice. Since we are using `docker/docker` anyway, we can also import `github.com/docker/docker/pkg/term`, which provides the below `GetWinsize` function which returns the current dimensions of terminal, just like @skanehira does.

```go
	winSize, err := term.GetWinsize(os.Stdout.Fd())
	if err != nil {
		os.Exit(1)
	}

	if winSize.Width == 0 || winsize.Height == 0 {
		signalCh := make(chan os.Signal, 1)
		signal.Notify(signalCh, syscall.SIGWINCH, syscall.SIGINT)
		<-signalCh
	}
```

I haven't tested it though (waiting for go modules support in lazydocker), but I think it should be okay after minor tweaks. Just put this code before initializing Gui I guess.
Could someone test it @jesseduffield @qdm12 ? Or maybe someone has a better solution?

README.md
dawidd6: `$HOME/.config/jesseduffield/lazydocker/config.yml`

Dockerfile
qdm12: There is a fix, using this script as entrypoint:

```sh
#!/bin/sh

resize > /dev/null
lazydocker $@
```

But it would indeed be cleaner to have it work in Go, so we could even have a scratch based container.

Dockerfile
```diff
@@ -7,16 +7,17 @@ FROM ${BASE_IMAGE_BUILDER}:${GO_VERSION}-alpine${ALPINE_VERSION} AS builder
 ARG GOARCH=amd64
 ARG GOARM
 ARG VERSION=0.2.4
```
jesseduffield: could we pass this in as an argument as well? Something like `--build-arg VERSION=$(git describe --abbrev=0 --tag)`

jesseduffield:
as for the resize thing, perhaps we can something like this in app.go:
```
func (app *App) Run() error {
	// before we do anything, we need to check that we have some window space available
	for {
		width, _, err := terminal.GetSize(int(os.Stdin.Fd()))
		if err != nil {
			return err
		}
		if width == 0 {
			time.Sleep(time.Millisecond * 50)
		}
		break
	}

	err := app.Gui.RunWithSubprocesses()

	return err
}
```

I'm in the middle of a long-running migration from dep to go modules right now so I can't actually test that locally :P

Dockerfile
qdm12: well yes you can, and it will override the `ARG VERSION=0.2.4`, the one set here is just the default one if none is given. I can set it to nothing if you prefer?

jesseduffield: I reckon we should leave VERSION to be blank if not explicitly passed, so that we don't incorrectly think that a person was using version 0.2.4 when they raise an issue.

The automated build on docker hub sounds like a good idea, and I'm sure it's possible with circle CI. I'll make an issue for that

jesseduffield: https://github.com/jesseduffield/lazydocker/issues/66

hooks/build
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
```
dawidd6: Is this script made for CI pipeline?

qdm12: It's picked up by Docker Hub automated build, see [this page](https://docs.docker.com/docker-hub/builds/advanced/) about it. But it won't be picked up by Circle CI.

dawidd6: Oh, TIL, thanks.

README.md
dawidd6: Well, personally I would keep two variants for user to choose. Looks like `docker build` is similar to `go install (locally cloned repo) | go get (pulling and installing)`.

qdm12: Yeah that's how it is now on my branch. However, with the Docker Hub build, it will pull the Docker image by itself when a `docker run` command is invoked, if the image does not exist so there is no need to build or even pull the image (except to update). I will leave the docker build instructions somewhere below the first few instructions, for developers/contributors because the average user should not need it 👍

Dockerfile
ibnesayeed: It is worth noting that things like `--build-arg VERSION=$(git describe --abbrev=0 --tag)` (or any other value that relies on certain `git` commands to extract a value) won't work when the build context is the Git URL and not built from a local clone or the repo. This is not something that is suggested in the modified README of this PR.

qdm12: Hi there! That seems like an odd bug from Docker, are you sure about it?

I'm waiting for @jesseduffield 's Docker Hub username and Docker Hub image name so that I update (push) a new readme in which
- the user just pulls and run the Docker image from Docker Hub, there is required building (except for ARM)
- the developer/contributor git clones and then docker builds locally, with its build arguments such as `VERSION`, it's cleaner than building from the `.git` URL.

qdm12: Actually that's a fair point, @jesseduffield  maybe having default values for build arguments (such as `VERSION=0.2.4`) would be better, if someone builds from the git repository?

pkg/app/app.go
```diff
@@ -51,8 +55,18 @@ func NewApp(config *config.AppConfig) (*App, error) {
...
+	// before we do anything, we need to check that we have some window space available
+	for {
+		width, _, err := terminal.GetSize(int(os.Stdin.Fd()))
+		if err != nil {
+			return err
+		}
+		if width == 0 {
+			time.Sleep(time.Millisecond * 50)
+		}
+		break
```
golangcibot: SA4004: the surrounding loop is unconditionally terminated (from `staticcheck`)

README.md
qdm12: We could use a volume as well so there is persistence without having to bind mount, but that's always a bit confusing on how to use for newcomers too. What do you think?

pkg/app/app.go
```diff
@@ -51,8 +55,17 @@ func NewApp(config *config.AppConfig) (*App, error) {
...
-
+	// before we do anything, we need to check that we have some window space available
+	var err error
+	width := 0
+	for width == 0 {
+		width, _, err = terminal.GetSize(int(os.Stdin.Fd()))
+		if err != nil {
+			return err
+		}
+		time.Sleep(time.Millisecond * 50)
```
ibnesayeed: I am afraid, this may cause an infinite loop unless we are sure that there can be no way that the width will remain `0` indefinitely. One rudimentary way to mitigate this issue is to keep a loop counter, which, when exceeds a threshold (say `20`) the app exits with a message like `Unable to initialize the window`.

README.md
dawidd6: `$HOME/.config/jesseduffield/lazydocker` would be best, I think.

pkg/app/app.go
qdm12: Good point. On the other hand, the container crashes if the window is of zero size, so a perhaps better way would be to log to stdout that it is waiting for the window to be created?

By the way, can lazydocker be ran without a container and with a zero sized window? If it is not possible either, the above solution seems like a good fit.

ibnesayeed: In my opinion, such logs should go to `STDERR` not `STDOUT`.

qdm12: I've added that in [the latest commit](https://github.com/jesseduffield/lazydocker/pull/36/commits/29ad43c4a4fdbc09d8085c68c820fbde8aa473a2) where it prints to stderr only when the terminal width is still zero after a first 50 milliseconds sleep.

pkg/app/app.go
```diff
@@ -55,18 +56,32 @@ func NewApp(config *config.AppConfig) (*App, error) {
...
 			return err
 		}
-		time.Sleep(time.Millisecond * 50)
+		if width != 0 {
+			return nil
+		}
+		count++
+		time.Sleep(time.Millisecond * 50 * time.Duration(count))
+		if count > 1 {
+			fmt.Fprintf(os.Stderr, "waited %s for available terminal space\n", time.Since(tStart))
```
ibnesayeed: This message is not very friendly for someone who might not know what exactly "terminal space" means. Also, the first letter of the first word should be in the upper case.

jesseduffield: I actually wonder if we need to print this more than once. At 50 milliseconds we'll get 20 of these lines per second, which if we're for some reason waiting a while, will be printing way more lines than we need. I don't actually see an issue with just printing once 'Waiting for available window space to render panels" after the width comes back as 0 the first time.

ibnesayeed: If we are going to print it once then we must not include the wait duration in the message.

qdm12: Yes I agree with all your points. I will simply do a print to stderr if the terminal space has zero width after 50 milliseconds (which should not happen)

jesseduffield: sounds good to me :)

Dockerfile
jesseduffield: even so, any static values we give for those will be quickly outdated. Do you mean we should make it try to grab the current version when building it locally? I wonder how many people would opt to do that as opposed to just pulling the image from docker hub.

BTW the docker hub username is `lazyteam` and the repository is `lazydocker`

pkg/app/app.go
dawidd6: I propose this solution:
```diff
diff --git a/pkg/app/app.go b/pkg/app/app.go
index 4f84543..49d6478 100644
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -4,7 +4,9 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"os/signal"
 	"strings"
+	"syscall"
 	"time"
 
 	"golang.org/x/crypto/ssh/terminal"
@@ -66,21 +68,23 @@ func (app *App) Run() error {
 
 func waitForTerminalSpace() error {
 	// before we do anything, we need to check that we have some window space available
-	tStart := time.Now()
-	count := 0
-	for {
-		width, _, err := terminal.GetSize(int(os.Stdin.Fd()))
-		if err != nil {
-			return err
-		}
-		if width != 0 {
-			return nil
-		}
-		count++
-		time.Sleep(time.Millisecond * 50 * time.Duration(count))
-		if count > 1 {
-			fmt.Fprintf(os.Stderr, "waited %s for available terminal space\n", time.Since(tStart))
-		}
+	width, height, err := terminal.GetSize(int(os.Stdin.Fd()))
+	if err != nil {
+		return err
+	}
+	if width != 0 && height != 0 {
+		return nil
+	}
+
+	winch := make(chan os.Signal, 1)
+	signal.Notify(winch, syscall.SIGWINCH)
+	timeout := time.Second * 5
+
+	select {
+	case <-winch:
+		return nil
+	case <-time.After(timeout):
+		return fmt.Errorf("waited %s for available terminal space\n", timeout)
 	}
 }
 ```

Comment (qdm12) [reply pkg/app/app.go]: That's an elegant solution, adopting it ! But I think a delay of 1 second should be enough for any device, instead of 5. Also, is `syscall.SIGWINCH` working across all platforms? I was thinking to add a `terminal.GetSize` check again after the timeout in case SIGWINCH did not trigger. But if it works everywhere, then I'm happy as it is now.

Comment (dawidd6) [reply pkg/app/app.go]: Looking at `go` source code it is present for Darwin+Linux on all architectures that we release `lazydocker` for.

File: pkg/app/app.go [hunk 9]
```diff
@@ -51,11 +58,33 @@ func NewApp(config *config.AppConfig) (*App, error) {
...
+	// before we do anything, we need to check that we have some window space available
+	width, height, err := terminal.GetSize(int(os.Stdin.Fd()))
+	if err != nil {
+		return err
+	}
+	if width > 0 && height > 0 {
+		return nil
+	}
+	winch := make(chan os.Signal)
+	signal.Notify(winch, syscall.SIGWINCH)
```
golangcibot: SA1017: the channel used with signal.Notify should be buffered (from `staticcheck`)

pkg/app/app.go
dawidd6: SIGWINCH is not present on windows tho, but seems like @jesseduffield is looking to support only WSL here https://github.com/jesseduffield/lazydocker/issues/29 and I'm in favor of this decision.

qdm12: Yes anyway the terminal size issue should only occur in Docker, which is running in Linux, so it should be fine.

dawidd6:
I think we may also need to either mount `docker` binary in container or just install it in it, cause it is needed to get some info (logs, cpu usage) right now.

ibnesayeed:
> I think we may also need to either mount docker binary in container or just install it in it, cause it is needed to get some info (logs, cpu usage) right now.

Since we will already be mounting the socket file, this perhaps can be done better by using APIs rather the a CLI binary.

qdm12:
Yes I agree, using the docker socket would be better. Especially that the docker binary can be in different locations depending on the host. 
Couldn't this also allow to have a remote Docker host connected?

qdm12:
I think for now I'll just copy in some docker static binary/binaries next to lazydocker in the scratch docker image so it can use them. But which binaries are needed? Only `docker` or also `dockerd` for example?

qdm12:
Actually after some thoughts, it's better to bind mount the Docker binary as its version might be different from the host, potentially causing issues.

pkg/app/app.go
```diff
@@ -51,11 +58,33 @@ func NewApp(config *config.AppConfig) (*App, error) {
...
+	if width > 0 && height > 0 {
+		return nil
+	}
+	winch := make(chan os.Signal, 1)
+	signal.Notify(winch, syscall.SIGWINCH)
+	select {
+	case <-winch:
+		return nil
+	case <-time.After(time.Second):
+		return fmt.Errorf("there is no available terminal space")
```
ibnesayeed: Should this string be localization-friendly? Should the first letter be in upper case?

ibnesayeed:
> Actually after some thoughts, it's better to bind mount the Docker binary as its version might be different from the host, potentially causing issues.

Bind mount for the binary is not ideal unless the binary on the host is static and can work cross-OS. I have created a new ticket #109 to use the API instead of the CLI binary.

pkg/app/app.go
qdm12: I should be removed soon anyway as gocui is getting fixed I think.

jesseduffield: the gocui fix has been merged into master now @qdm12 so feel free to delete it from this branch :)

jesseduffield:
I forgot where we were up to with this. If I recall correctly, it was ready to go? I think there might have been some discussion on another issue but I can't find it now

qdm12:
I need this [last question resolved](https://github.com/jesseduffield/lazydocker/issues/28#issuecomment-509020969)

Also now that you use go modules, I need to changed the go binary build stage as it should be different now. I will do that the coming 2 days.

pascalandy:
I'll be glad to test if you need help :)

qdm12:
Hi all,

I tested my latest Dockerfile with the lastest master code (using modules) and it fails with:

```sh
go: finding github.com/davecgh/go-spew v1.1.1
go: finding github.com/client9/misspell v0.3.4
go: finding golang.org/x/sys v0.0.0-20180905080454-ebe1bf3edb33
go: finding github.com/onsi/ginkgo v1.6.0
go: error loading module requirements
The command '/bin/sh -c go mod download' returned a non-zero code: 1
```

during my `go mod download` stage.

The full Dockerfile is:

```Dockerfile
ARG BASE_IMAGE_BUILDER=golang
ARG ALPINE_VERSION=3.10
ARG GO_VERSION=1.12.6

FROM ${BASE_IMAGE_BUILDER}:${GO_VERSION}-alpine${ALPINE_VERSION} AS builder
ARG GOARCH=amd64
ARG GOARM
RUN apk --update -q --progress add git ca-certificates
RUN go get -u -v golang.org/x/vgo 2> /dev/null
WORKDIR /tmp/gobuild
COPY go.mod go.sum ./
RUN go mod download
COPY ./ .
RUN CGO_ENABLED=0 GOOS=linux GOARCH=${GOARCH} GOARM=${GOARM} go build -a -ldflags="-s -w \
    -X main.commit=${VCS_REF} \
    -X main.version=${VERSION} \
    -X main.buildSource=Docker"

FROM scratch
ARG BUILD_DATE
ARG VCS_REF
ARG VERSION
LABEL org.label-schema.schema-version="1.0.0-rc1" \
    maintainer="jessedduffield@gmail.com" \
    org.label-schema.build-date=$BUILD_DATE \
    org.label-schema.vcs-ref=$VCS_REF \
    org.label-schema.url="https://github.com/jesseduffield/lazydocker" \
    org.label-schema.vcs-description="The lazier way to manage everything docker" \
    org.label-schema.docker.cmd="docker run -it -v /var/run/docker.sock:/var/run/docker.sock lazydocker" \
    org.label-schema.version=${VERSION}
ENTRYPOINT [ "/lazydocker" ]
VOLUME [ "/.config/jesseduffield/lazydocker" ]
COPY --from=builder /tmp/gobuild/lazydocker /lazydocker
```

Any idea why is this? Thanks!

Dockerfile
```diff
@@ -28,4 +42,5 @@ LABEL org.label-schema.schema-version="1.0.0-rc1" \
     org.label-schema.version=${VERSION}
 ENTRYPOINT [ "/lazydocker" ]
 VOLUME [ "/.config/jesseduffield/lazydocker" ]
-COPY --from=builder /go/src/github.com/jesseduffield/lazydocker/lazydocker /lazydocker
\ No newline at end of file
+ENV PATH=/
+COPY --from=docker-builder /go/src/github.com/docker/cli/build/docker /docker
```
ibnesayeed: Perhaps it would be better to place `docker` binary at `/usr/bin/docker`, which would not require overwriting `PATH` environment variable (hence one less layer).

Dockerfile
```diff
@@ -1,15 +1,47 @@
...
+    maintainer="jessedduffield@gmail.com" \
+    org.label-schema.build-date=$BUILD_DATE \
+    org.label-schema.vcs-ref=$VCS_REF \
+    org.label-schema.url="https://github.com/jesseduffield/lazydocker" \
+    org.label-schema.vcs-description="The lazier way to manage everything docker" \
+    org.label-schema.docker.cmd="docker run -it -v /var/run/docker.sock:/var/run/docker.sock lazydocker" \
+    org.label-schema.version=${VERSION}
+ENTRYPOINT [ "/bin/lazydocker" ]
+VOLUME [ "/.config/jesseduffield/lazydocker" ]
+ENV PATH=/bin
```
ibnesayeed: Since binaries are placed in one of the default `PATH` locations, this is not needed anymore.

```suggestion
```

Dockerfile
```diff
@@ -1,15 +1,47 @@
...
+ARG VCS_REF
+ARG VERSION
+LABEL org.label-schema.schema-version="1.0.0-rc1" \
+    maintainer="jessedduffield@gmail.com" \
+    org.label-schema.build-date=$BUILD_DATE \
+    org.label-schema.vcs-ref=$VCS_REF \
+    org.label-schema.url="https://github.com/jesseduffield/lazydocker" \
+    org.label-schema.vcs-description="The lazier way to manage everything docker" \
+    org.label-schema.docker.cmd="docker run -it -v /var/run/docker.sock:/var/run/docker.sock lazydocker" \
+    org.label-schema.version=${VERSION}
```
ibnesayeed: [Label Schema](http://label-schema.org/rc1/) is deprecated in favor of [OCI Image Spec](https://github.com/opencontainers/image-spec/blob/master/annotations.md).

dawidd6:
@qdm12 

> ```
RUN go get -u -v golang.org/x/vgo 2> /dev/null
```

Why is this needed? Just install `go` with `apk`.

Dockerfile
qdm12: I think `PATH` is not defined on Scratch images. I defined it (in the next commits) as `/bin` and placed the binaries there.

Dockerfile
qdm12: Good to know I'll update it! And my other images :smile:  !

qdm12:
It's just to make the build faster, so that dependencies are not downloaded at every build. Do you think it's possible to `go mod download` and use the vendor directory at the same time?

For now I'll make it a one line build as you suggested but this is quite slow for development purposes.

Dockerfile
qdm12: Can anyone test that `PATH` is actually defined on a scratch based image? Probably using a go program I think.

I believe it's not so I will keep this ENV line for now.

qdm12: Never mind, you were right, `PATH` is already defined. I've applied your suggestion! Thanks!

dawidd6:
But you don't need to download dependencies as there are already available offline in `vendor/` and I don't know how can that slow down the build.

qdm12:
Hi, that's great then, less work for me :smile: It would be eventually interesting to get rid of the vendor directory and have only git commit references in the go.mod file, or am I missing something out? Thanks.

dawidd6:
@jesseduffield likes vendoring, cause it makes it easier to quickly make changes to dependencies and see the results. Plus also with vendoring it is enough to just clone one repo and then build, while without vendoring there is also a need to download dependencies from different sources before building.

So, in conclusion, I think that vendoring will stay here for a while.

dawidd6:
I just tried the newest iteration of `Dockerfile` and it seems like logs, stats, attaching and other stuff handled by docker CLI binary are still not working. I tried using alpine image instead of the scratch one, as I thought it could be an issue with PATH, but it wasn't the case. Dunno really what to do with that.

qdm12:
I had my CPU stats working on my side. I'll try attaching to containers as well. Are you sure you are trying the last Dockerfile with the current master branch?

dawidd6:
Yep, i just merged this branch from your fork with latest upstream master and I don't get those stats.

qdm12:
Hi all, sorry for the long delay.

I get the stats, config and Top. Screenshot below:

image

But I don't get the logs. What are your thoughts on this? I cannot figure out why it does not work.

dawidd6:
Me neither. I get the same result.

I'd suggest merging this PR already as it is very outdated comparing to current master (go modules) and open another PR for fixing remaining issues.

qdm12:
Yes I wanted to suggest the same thing but didn't want to be rude.

It will still be much better than the current Dockerfile implementation. And, as you pointed out, it'll be easier to develop from a new branch as copying back and forth between repositories (one with modules, one without) isn't great.

dawidd6:
It is done. Please open another PR.

tmardesen-greatamerica:
I am using the latest Go installation, and I do not have Top or Stats working (This is installed on windows server 2019), what is the work around? To use the dockerfile for Lazydocker?

Issue #11, opened 2019-06-30
Support the docker-compose -f flag


mcintyre94:
In our app we have multiple docker-compose files, eg. for running the app locally, and then subsets for functional/integration tests with various different configuration/volume adjustments. It'd be awesome to be able to do `ld -f docker-compose-functional-test.yml` and have it run within the context of that file. Would something like that make sense?

jesseduffield:
interestingly my own company does something similar. We have a program that you can basically use in place of docker-compose and you're able to do it from any directory and it will work. So I made setting the docker-compose command configurable. You just need to do have:

```
commandTemplates:
  dockerCompose: 'custom-compose'
```

In your case it would be `dockerCompose: 'docker-compose -f foo.yml`

That satisfies my use case where we only ever really need to be in a single context, but it seems like you would need to change your 'dockerCompose' config value depending on your current context. I think making it a command-line arg makes sense, though I'm not sure what the best approach would be.

One approach would be to pass in the entire docker-compose command like
```
ld --dc "docker-compose -f foo.yml -f bar.yml"
```

This is the most flexible because it supports adding pretty much anything you want into that command, and allows as many -f arguments as you want. It's also simply because it will just replace the default docker-compose command from your config. But it might be annoying to have to write 'docker-compose' all the time when you're not somebody who has their own custom docker-compose program.

So maybe it should be that if you supply -f flags, each one is just going to be tacked onto whatever the docker-compose command is from the config. And then we'll wait and see if anybody needs more flexibility than that.

What are your thoughts?

mcintyre94:
Our use case is always just one docker-compose file at a time, I didn't actually realise you could pass multiple! So for me `ld -f foo.yml` would be perfect as opposed to `ld --dc "docker compose -f foo.yml"`, and for the multiple files case multiple `-f` flags definitely sounds nicer if that works. I guess the other question would be whether there's any other `docker-compose` flags others would want to pass through though, since at some point it might be preferable to use the `--dc` approach than proxy multiple flags through.

jesseduffield:
If you wanna have a crack be my guest :)
[...]

PR #38, opened 2019-07-01
Migrate to flaggy

mcintyre94:
As discussed in https://github.com/jesseduffield/lazydocker/issues/11, migrating from Flag to Flaggy is a prerequisite for supporting slice arguments. This PR moves the existing code from Flag to Flaggy without any changes in behaviour or additional flags being added yet.

One thing to note is that Flaggy has a built-in version flag, which I'm using with the previous version output.

I've also added a Name/Description/Repo link to the help output - those are in a second commit in case they're not wanted/you want them changed.

New help output:

```
$ lazydocker --help
lazydocker - The lazier way to manage everything docker
https://github.com/jesseduffield/lazydocker

  Flags:
       --version  Displays the program version string.
    -h --help  Displays help with available flag, subcommand, and positional value parameters.
    -config --config  Print the current default config
    -d --debug  a boolean
```

Version output (on my machine):

```
$ lazydocker --version
Version: commit=, build date=, build source=unknown, version=unversioned, os=darwin, arch=amd64
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -44,9 +44,6 @@ func main() {
 		os.Exit(0)
 	}
 
-	// for now we're always in debug mode so we're not passing *debuggingFlag
-	debuggingFlag = true
-
 	appConfig, err := config.NewAppConfig("lazydocker", version, commit, date, buildSource, debuggingFlag)
 	if err != nil {
 		log.Fatal(err.Error())
--- a/main.go
+++ b/main.go
@@ -2,7 +2,6 @@ package main
 
 import (
 	"bytes"
-	"flag"
 	"fmt"
 	"log"
 	"os"
@@ -10,6 +9,7 @@ import (
 
 	"github.com/docker/docker/client"
 	"github.com/go-errors/errors"
+	"github.com/integrii/flaggy"
 	"github.com/jesseduffield/lazydocker/pkg/app"
 	"github.com/jesseduffield/lazydocker/pkg/config"
 	"github.com/jesseduffield/yaml"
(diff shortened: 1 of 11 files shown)


main.go
```diff
@@ -21,29 +21,33 @@ var (
...
-		fmt.Printf("commit=%s, build date=%s, build source=%s, version=%s, os=%s, arch=%s\n", commit, date, buildSource, version, runtime.GOOS, runtime.GOARCH)
-		os.Exit(0)
-	}
 
-	if *configFlag {
+	flaggy.SetName("lazydocker")
+	flaggy.SetDescription("The lazier way to manage everything docker")
+	flaggy.DefaultParser.AdditionalHelpPrepend = "https://github.com/jesseduffield/lazydocker"
+
+	flaggy.Bool(&configFlag, "config", "config", "Print the current default config")
```
dawidd6: I think that the shorthand for `config` should be `c`.

mcintyre94: Oops, thankyou!

dawidd6:
Sooo, are you going to add more flags, like `-f` within this pull request or will you make another one for this?

mcintyre94:
@dawidd6 If you're happy with these changes, shall I squash into one commit?

dawidd6:
Sure, can't see why not.

main.go
```diff
@@ -21,29 +21,33 @@ var (
...
+	if configFlag {
 		var buf bytes.Buffer
 		yaml.NewEncoder(&buf).Encode(config.GetDefaultConfig())
 		fmt.Printf("%v\n", buf.String())
 		os.Exit(0)
 	}
 
 	// for now we're always in debug mode so we're not passing *debuggingFlag
-	*debuggingFlag = true
+	debuggingFlag = true
```
jesseduffield: we can actually remove lines 47 and 48 because that was only there for my own personal convenience :P

mcintyre94: Good shout! :)

mcintyre94:
@jesseduffield Newbie go question! I ran `dep ensure`, and it also added `vendor/github.com/jesseduffield/lazydocker/` - is there any way to avoid that happening in my fork? I figure we don't want it here :)

dawidd6:
@mcintyre94 did you change some imports to point to your fork?

mcintyre94:
> @mcintyre94 did you change some imports to point to your fork?

Not that I can see - `git grep mcintyre94` is empty, and the only diff to what's currently pushed here is that I ran `dep ensure` so no import changes that aren't pushed.

mcintyre94:
I can remove it before I push, but just wondered if there was something I could do to avoid that happening next time :)

jesseduffield:
@mcintyre94 I'm actually not sure how to get around that issue. It might be a limitation of using dep with forks :P

jesseduffield:
At any rate, I'm happy with this PR so it's merge time!

PR #45, opened 2019-07-01
Add automated install-update steps for Linux

immanuelfodor:
Hi, I've added a small code snippet to be able to download and install/update the binary file without manual steps. Tested on Ubuntu and Manjaro Linux distros. Also improved the clickability of the release link.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,19 @@ brew install lazydocker
 
 ### Binary Release (Linux/OSX)
 
-You can download a binary release [here](https://github.com/jesseduffield/lazydocker/releases).
+You can manually download a binary release from [the release page](https://github.com/jesseduffield/lazydocker/releases).
+
+Automated install/update:
+
+```sh
+GITHUB_LATEST_VERSION=$(curl -L -s -H 'Accept: application/json' https://github.com/jesseduffield/lazydocker/releases/latest | sed -e 's/.*"tag_name":"\([^"]*\)".*/\1/')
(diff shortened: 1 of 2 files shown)


dawidd6:
This, to be fully automated would need to be in form of a script in file, so one could just pipe curl to bash (which isn't that good idea).

What I'm trying to say is that, if you really want to make this right, you should put theses commands in file and store it in https://github.com/jesseduffield/lazydocker/tree/master/scripts (probably). Then in README, you can write, that user should execute: `curl url_to_raw_script | bash`, to get the latest release.

immanuelfodor:
I'm not really a fan of piping into bash, to be honest.  
Should I create the script anyway, or we just call it semi-automatic way? :grinning:

dawidd6:
Well, the script anyway would be present in repository, so everyone interested would have the possibility to review it thoroughly before piping it to bash. Also the script seems to not be very complicated or long.

So, hmm, after all, I would vote for storing it in scripts/ directory.

immanuelfodor:
Okay, I think this is it.

immanuelfodor:
I've just seen the version check part: the `mv -f` command always rewrites the installed version, so you can use it to install or update the existing binary file.

immanuelfodor:
@jesseduffield what do you think, are we ready to merge? :)

jesseduffield:
Thanks for making this! I'm satisfied, what about you @dawidd6 ?

dawidd6:
@jesseduffield yep, should be fine.

jesseduffield:
mergin!

PR #54, opened 2019-07-01
Updates README.md with a better alias suggestion

icdevin:
`ld` is actually a pretty important command-line tool, so overwriting it with the suggested alias is not preferable. `lzd` is just my idea, but really this is just to notify you.

More on `ld`: https://linux.die.net/man/1/ld


jesseduffield:
agreed!

PR #55, opened 2019-07-01
Add -f compose file flag

### LINKED ISSUE #11: Support the docker-compose -f flag

mcintyre94:
Builds on #38 (first three commits are from there), closes #11 

Using Flaggy, adds the -f flag which allows specifying multiple docker-compose yml files which are passed into the `docker-compose` command. This allows lazydocker to support multiple arbitrary docker-compose files.

Also added some unit tests for this new code.

@markphip Does this address your use case too? :)

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -114,6 +114,14 @@
   version = "v0.3.7"
 
 [[projects]]
+  digest = "1:9cf73012dbe62f96cd8cdd210a1123a360401967a4530284dcd2990888747e3d"
+  name = "github.com/integrii/flaggy"
+  packages = ["."]
+  pruneopts = "NUT"
+  revision = "07ea7eb774047dab1bdcc7cad0f825fa419df47d"
+  version = "1.2.1"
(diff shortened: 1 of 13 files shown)


markphip:
If I can do something like this then I would say yes:

`lazydocker -f docker-compose.yml -f docker-compose-srv1.yml -f docker-compose-srv2.yml`

I was able to successfully use a custom-compose script but this would be better for my workflow.

mcintyre94:
That should be exactly what this adds :)


________________________________

markphip:
Thanks I built this PR locally and confirmed it works nicely.

jesseduffield:
another fantastic PR, and thanks for adding the tests! mergin

PR #60, opened 2019-07-02
Fix main panel focus bug

mjarkk:
**Problem**  
After second (mouse) click on main panel the `[`, `]` and `esc` stop working.  
This fixes that.


jesseduffield:
good pickup!

Issue #63, opened 2019-07-02
Barely visible texts in terminal running solorized-dark color scheme


den-is:
**Describe the bug**
Some texts are barely visible in terminal which is using solarized-dark color scheme. 
See attachment.

**Screenshots**
<img width="607" alt="Screen Shot 2019-07-02 at 12 40 13" src="https://user-images.githubusercontent.com/5287237/60497792-904e0280-9cc6-11e9-9855-25ecf7249981.png">

**Desktop (please complete the following information):**
 - macos 10.14.5
 - iTerm 3.2.6
```
lazydocker version
Version: commit=cf31c23fadcc717467318689fc073ecf1bed8031, build date=2019-07-02T00:10:43Z, build source=binaryRelease, version=0.3, os=darwin, arch=amd64
```

**Additional context**
similar issues with more screenshots if that helps 
https://github.com/derailed/k9s/issues/132
https://github.com/derailed/k9s/issues/197#issuecomment-496809205

mjarkk:
Totally agree,  
![image](https://user-images.githubusercontent.com/15320763/60498579-2aa64a00-9cb7-11e9-9b0a-87345210b106.png)

jesseduffield:
Yeah I've also seen this on a colleagues computer. It's such a shame because it looks fine on my terminal, but I'm not sure what replacement color I could use. Luckily we already have color-customisable panel borders but as for things like these 'none's:

```
![image](https://user-images.githubusercontent.com/8456633/60505381-f5294d00-9d06-11e9-9651-3a80e12f2c01.png)

```

What might be a better option than black? I want them to look kind of out-of-the-way

jesseduffield:
blue doesn't look too bad:
![image](https://user-images.githubusercontent.com/8456633/60506282-e04db900-9d08-11e9-9040-f37744193b9c.png)

![image](https://user-images.githubusercontent.com/8456633/60506294-e479d680-9d08-11e9-9e94-ffe04b97e763.png)

Blue is usually the darkest of the colors so I think it suits. Just means now we'll need a different color for the container state of 'exited (0)' (I made it blue to distinguish a good exit from a bad exit). Maybe Yellow?

jesseduffield:
here's exited(0) in yellow:
![image](https://user-images.githubusercontent.com/8456633/60506597-80a3dd80-9d09-11e9-8ee9-528bc6d1d66a.png)

I think it looks appropriate

jesseduffield:
[...]

PR #65, opened 2019-07-02
improve readability of colours

jesseduffield:
fixes https://github.com/jesseduffield/lazydocker/issues/63

### DIFF
--- a/pkg/commands/container.go
+++ b/pkg/commands/container.go
@@ -294,7 +294,7 @@ func (c *Container) GetColor() color.Attribute {
 	switch c.Container.State {
 	case "exited":
 		if c.Details.State.ExitCode == 0 {
-			return color.FgBlue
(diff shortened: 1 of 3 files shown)


mjarkk:
Isn't is a good idea to make this user configurable?

jesseduffield:
You're probably right. I just wonder how configurable we want to go. Do we say want something like:
```
gui:
  theme:
    servicePanel:
      status:
        none: blue
        running: green
        exitedZero: yellow
        exitedNonZero: red
```

mjarkk:
Looks good to me.  
For the people that bother changing the colors they probably want to have full control over all colors so would not be a bad idea to give them full access to all colors.

jesseduffield:
agreed. I think I'll merge this PR, which will fix the immediate issue, then put up an issue to allow full customisation of colours

PR #69, opened 2019-07-02
Move to Go Modules

tommyknows:
I wanted to install this project through `go get`, which failed because I have go modules turned on by default.

This PR moves to modules which become (fully) enabled by default in Go 1.13.
As Go 1.12 is already a hard requirement, this should not break anything.

Note: I had to specify the exact commit for the `github.com/jesseduffield/yaml` project, because without any version specified, it clones `v2.1.0`.
Reason for this is, I suspect, that `v2.1.0` was the last tag without a `go.mod` file, and newer tags (e.g. >= `v2.1.1`) are "incompatible" because the import path in the `go.mod` is `gopkg.in/yaml.v2`.

I'd recommend tagging the yaml-repo and then executing go-get to fetch the newest tag with the correct import path 🙂


dawidd6:
Heh, as a matter of fact we were just trying to make a move to Go modules with @jesseduffield 
Good job. Saved us some time. Did you tested it fully (get, build, install, test) inside or outside GOPATH?

tommyknows:
I did not - reason for this is because I'd need to rewrite all the imports to my repo-URL.
(I did rewrite it and test that, which worked, but I can't really test it with the original import paths)

Possibility would be to merge this into another branch than master and then to test from there

dawidd6:
I see.
Actually there is already a branch made for that:
https://github.com/jesseduffield/lazydocker/tree/go-modules

Nevertheless, we should wait for repository owner, to push it and fully test it.

jesseduffield:
PR is up: https://github.com/jesseduffield/lazydocker/pull/74 but I have no idea what I'm doing! How would we go about testing this?

mikhailswift:
@jesseduffield Easiest way to test go mod is to clone the project to a folder not in your `$GOPATH` and run `go build`.

You'll likely want to delete your `vendor` directory.  Go Mod only copies folders with .go files to your vendor directory from it's package cache.  This can cause problems if any of your dependencies are cgo dependencies.

Go mod will automatically detect if you're in your $GOPATH and if not enable go mods and download dependencies for your project on any `go` command that requires it.  If you are in your `$GOPATH`, you can set `GO111MODULES` to `on`.

jesseduffield:
go modules does support using a vendor directory thought right? I have a strong inclination to hold onto it, given how often I currently make use of it (and it lets me make some random changes across various packages, then stash those changes when I switch to working on something else, so I can then unstash them and continue later).

mikhailswift:
It does kind of support the vendor directory. I edited my comment above but you can set mod=vendor in your go flags and it will operate off the vendor directory. 

I’m not sure if it’ll nuke your vendor directory and replace it with stuff from its own package cache. It might. go mod vendor will populate the vendor directory from your mod files. Then go build —mod=vendor will build from your vendor dir. They are looking to deprecate the vendor dir though.

jesseduffield:
closing this PR to keep the conversion in https://github.com/jesseduffield/lazydocker/pull/74 but thanks again @tommyknows for making this :)

PR #75, opened 2019-07-03
Launch OpenFile command in sub process

### LINKED ISSUE #71: Hitting `o` to open config hangs
**Describe the bug**
Hitting `o` while on the status pane to view my config hangs the app. 

**To Reproduce**
1. Go to status pane
2. Hit `o`

**Expected behaviour**
Config displays

**Screenshots**

 - OS: Arch Linux
 - Lazydocker Version 0.3.0

**Additional context**
nvim seems to be the app xdg-open chooses to view the config.yml file.  I can run `xdg-open ~/.config/jesseduffield/lazydocker/config.yml` (which seems to be the command lazydocker runs) in a shell to open the config file in nvim successfully.  I can also hit `e` in lazydocker to edit the config and it opens for edit successfully.  When I'm not on work hours I can play around with a development build and see if I can determine what is happening.

mikhailswift:
Fixes #71 

My `xdg-open` opens vim.  With this change and setting my .config file to have:

```
oS:
  openCommand: xdg-open {{filename}}
```
I can now open the config file from lazydocker's TUI.

Can someone on Windows/OSX try this out?  Is this change a good idea?  I'm unsure how this will affect user's whose `xdg-open` doesn't open a terminal program or another operating system's use case.  Is removing the redirect to >dev/null in the default open command appropriate?

### DIFF
--- a/pkg/commands/os.go
+++ b/pkg/commands/os.go
@@ -128,15 +128,14 @@ func sanitisedCommandOutput(output []byte, err error) (string, error) {
 }
 
 // OpenFile opens a file with the given
-func (c *OSCommand) OpenFile(filename string) error {
+func (c *OSCommand) OpenFile(filename string) (*exec.Cmd, error) {
 	commandTemplate := c.Config.UserConfig.OS.OpenCommand
 	templateValues := map[string]string{
 		"filename": c.Quote(filename),
 	}
 
-	command := utils.ResolvePlaceholderString(commandTemplate, templateValues)
-	err := c.RunCommand(command)
-	return err
+	command := c.ExecutableFromString(utils.ResolvePlaceholderString(commandTemplate, templateValues))
+	return command, nil
 }
 
 // OpenLink opens a file with the given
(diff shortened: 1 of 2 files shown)


mikhailswift:
I tried this branch on my macbook and it works, but it does result in the "Press enter to return to lazydocker" prompt.

jesseduffield:
I think that you might be better off making use of the 'e' keybinding on the status panel to edit the config, which ends up calling this code:
```
func (c *OSCommand) EditFile(filename string) (*exec.Cmd, error) {
	editor, _ := c.getGlobalGitConfig("core.editor")

	if editor == "" {
		editor = c.getenv("VISUAL")
	}
	if editor == "" {
		editor = c.getenv("EDITOR")
	}
	if editor == "" {
		if err := c.RunCommand("which vi"); err == nil {
			editor = "vi"
		}
	}
	if editor == "" {
		return nil, errors.New("No editor defined in $VISUAL, $EDITOR, or git config")
	}

	return c.PrepareSubProcess(editor, filename), nil
}
```

(I am open to removing that core.editor part, it's a vestige of lazygit.)

Would that satisfy your use case?

mikhailswift:
The main issue I have is that `o` currently hard locks my instance of lazy docker.  Since this is more of a configuration issue on my end I can just set my `openFile` config to a no-op command.

Issue #41, opened 2019-07-01
Does not fully work on Kubuntu 18.04


Janealter:
Services and standalone containers are not displayed.

dawidd6:
Docker version installed on your machine?
Compose version installed on your machine?
Lazydocker installation method and version as well on your machine?

Please provide some more information.

piotrmacha:
I had similar problem running lazydocker on WSL with docker-compose
```
docker-compose version 1.17.1, build unknown
```
Updating docker-compose to newest version solved the problem
```
docker-compose version 1.24.0, build 0aa59064
```

You can find instructions for getting the newest version here on https://docs.docker.com/compose/install/#install-compose-on-linux-systems

arpadszasz:
I can confirm the solution provided by @piotrmacha works for me on:
* Ubuntu 19.04 (disco)
* Docker version 18.09.5, build e8ff056 (from Ubuntu universe repo)
* **updated** docker-compose version 1.24.0, build 0aa59064

Thanks.

markphip:
Had a colleague also run into this. Would be good to identify the minimum version required in the README.md.  This is what I have on MacOS where it works fine for me: `docker-compose version 1.23.2, build 1110ad01`

cmpilato:
I'm the colleague @markphip mentioned, and can confirm that upgrading docker-compose to 1.24 solved the issue.

PR #77, opened 2019-07-03
Mention requirements in README

dawidd6:
I think this closes: #41

Also Minimum required Go version seems to be 1.8. Checked that in Docker containers.


jesseduffield:
I don't even have go 1.8; why do we need such a recent version?

dawidd6:
Go getting with 1.6:
```
package context: unrecognized import path "context" (import path does not begin with hostname)
```

jesseduffield:
thanks @dawidd6 I just realised that 1.8 < 1.12 so my concern was unfounded :)

jesseduffield:
thanks :)

PR #85, opened 2019-07-03
Map the aarch64 platform to the arm64 binary

### LINKED ISSUE #83: aarch64 -> amd64 binary
**Describe the bug**
using `curl https://raw.githubusercontent.com/jesseduffield/lazydocker/master/scripts/install_update_linux.sh | bash` command fails is the system arch is `aarch64`

**To Reproduce**
1. Go to a machine with `aarch64` architecture
2. Execute `curl https://raw.githubusercontent.com/jesseduffield/lazydocker/master/scripts/install_update_linux.sh | bash`
3. See error: `ERROR 404: Not Found.`

**Expected behaviour**
Installation completed with `lazydocker_0.3_Linux_arm64.tar.gz` instead of non-existent `lazydocker_0.3_Linux_aarch64.tar.gz` installed

**Error log**
```
--2019-07-03 13:24:54--  https://github.com/jesseduffield/lazydocker/releases/download/v0.3/lazydocker_0.3_Linux_aarch64.tar.gz
Resolving github.com (github.com)... 140.82.118.4
Connecting to github.com (github.com)|140.82.118.4|:443... connected.
HTTP request sent, awaiting response... 404 Not Found
2019-07-03 13:24:54 ERROR 404: Not Found.


gzip: stdin: unexpected end of file
tar: Child returned status 1
tar: Error is not recoverable: exiting now
mv: cannot stat 'lazydocker': No such file or directory
```

**Additional context**
It would be nice to map `lazydocker_0.3_Linux_arm64.tar.gz` binary to `lazydocker_0.3_Linux_aarch64.tar.gz`

immanuelfodor:
Fixes #83

The solution is to check if `uname -m` is `aarch64`, then it should be `arm64` in the URL.

If we need to map more platforms later, we can introduce a simple `case...esac` check to handle the different outputs of `uname -m`.


dawidd6:
I think that also i386 should be mapped to 32-bit.

immanuelfodor:
Wouldn't it be easier to rename the binaries then? Just askin' :grinning:

dawidd6:
Yea, I'm thinking about that too. `32-bit` seems weird IMO.

dawidd6:
https://github.com/jesseduffield/lazydocker/pull/86

immanuelfodor:
Okay, so the aarch64 mapping is done by my `sed`, and the i386 is mapped by the renamed binary, correct? I can't test your modification as I don't have a 32-bit machine.

dawidd6:
`i386` is `x86` now

immanuelfodor:
Okay, here it is.

dawidd6:
On my Raspberry Pi 2, I get: `armv7l`

I think it would be better to do a `case` now, like:
```sh
case "$(uname -m)"; in
  i386|i686) arch=x86 ;;
  *armv6*) arch=armv6 ;;
  *armv7*) arch=armv7 ;;
  *aarch64*) arch=arm64 ;;
esac
```
it's just a suggestion, the above code may not work, I written it by hand quickly, but I think you get what I want to say :)

immanuelfodor:
Funny that I even suggested using a case statement on the top here, and then went for adding another sed rule 😀 I tested the change on a smart phone which was aarch64 as it turned out.

dawidd6:
Yea, I saw hehe. Looks good. Merging.

Issue #20, opened 2019-06-30
Any plans on making a snappy packages? (Question)


titus-anromedonn:
Love the project, this is awesome, great job!

I was wondering, do you have any plans on making a snappy package?

dawidd6:
If there is such demand, I could do it by directly contributing `snapcraft.yaml` to this repo.

But I would better like not to be the one who uploads to Snap store.
Definitely would be the best if the uploader was the original creator of software. @jesseduffield

titus-anromedonn:
> If there is such demand, I could do it by directly contributing `snapcraft.yaml` to this repo.
> 
> But I would better like not to be the one who uploads to Snap store.
> Definitely would be the best if the uploader was the original creator of software. @jesseduffield
> 
> NOTE for Jesse: you already have a launchpad account so there is simply no need to make another account and Canonical provides this [service](https://build.snapcraft.io) for automatic builds straight from Github repo.

Agreed. Personally, I would prefer to have the snappy package available for download instead of having to manually download the linux binary myself, but that's just me being lazy!

jesseduffield:
I'm happy to do this if I can get some guidance on what needs to be done :P

dawidd6:
I'll try to tackle this today.

dawidd6:
Seems like you are using goreleaser and it has snapcraft support:
https://goreleaser.com/customization/#Snapcraft

Maybe we should do it this way?

dawidd6:
I made two variants:
- added snapcraft.yaml
- modified .goreleaser.yml

https://github.com/jesseduffield/lazydocker/compare/master...dawidd6:snapcraft

PR #86, opened 2019-07-03
Update Goreleaser

dawidd6:
### Already done:
- moved env from local context to global
- defined separate build for snap
- build for armv6 and armv7
- fixed some deprecations:
  - archive -> archives
  - brew -> brews
- replaced 386 with x86 (for consistency with x86_64)
- added snapcraft (strict confinement)

### To be done:
- container logs cannot be obtained, cause there is no `docker` command in snap
- same problem will be with `docker-compose`

### Notes:
I managed to make it `strict` confined by defining `docker-support` plug.

So user after install would need to manually execute:
`snap connect lazydocker:docker-support`
or
`snap connect lazydocker:docker` if Docker is installed as snap I suppose.

If someone wants to test this, run:
```sh
goreleaser --skip-publish --snapshot --skip-sign --rm-dist
```

closes: #20

### Useful links for @jesseduffield :
https://docs.snapcraft.io/registering-your-app-name

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -114,8 +114,22 @@ snapcrafts:
     # permissions for strict snaps can be declared as `plugs` for the app, which
     # are explained later. More info about confinement here:
     # https://snapcraft.io/docs/reference/confinement
-    confinement: classic
+    confinement: strict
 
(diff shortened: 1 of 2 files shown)


jesseduffield:
LGTM

Issue #40, opened 2019-07-01
Feature request: show container names


mudasar187:
Hi, i'v tested out lazydocker today and i love it!
A feature to have is to have names on the containers where the `Services` window is.

I have a `docker-compose.yml` file:

`version: '3.7'

services:
  mongodb:
    image: mongo
    container_name: "db_app"
    ports:
      - "27017:27017"

  frontend:
    image: quizgame/webclient
    container_name: "web_app"
    build:
      context: .
      dockerfile: Dockerfile
    ports:
      - "8080:3000"
    environment:
      - NODE_ENV=docker
    depends_on:
      - mongodb`

In `Services` window, there is to running containers.

`running frontend 0.00%
  running mongodb 0.35%`

It should be:

`running frontend (web_app) 0.00%
  running mongodb (db_app) 0.35%`

Love this tool!

jesseduffield:
Sounds like a reasonable request. Given that there if often not much horizontal space, I'm inclined to make this a user configurable thing that defaults to false. Something like:
```
gui:
  showContainerName: true
```
What are your thoughts on that approach?

As for implementation, It would be fairly trivial to write the code, it would just be a matter of changing the GetDisplayStrings() function in `pkg/commands/service.go`. I'm happy to label this as Good First Issue and see if anybody wants to pick it up (I'm happy to lend a hand) :)

mudasar187:
Nice to hear :)

My thoughts is, make it default:

```
gui:
    showContainerName: true
```

And if someone dont want the container name then they can set it to `false`.
If if there is no more space in horizontal then you can show like these examples:

`running thisIsNameOfService (..) 0.00%` (service name is long then add (..) to container name
same on the other way.

And when you click on this service it will show a header on the left windows where `Logs, Stats, Container Config, Top` is and show the container name on the header.

pvande:
My $0.02: I don't really want or need to see the container name in the sidebar (in the header of the main panel would be great though!) **unless** the service has a dynamic container_name.  Knowing what directory lazydocker is running in would be more useful.

My primary use case is in development, where I run a few distinct projects that share service names.  My services have static container names to make linking them together easier, and I'm never running a service as more than a single instance.  If/when I cannot supply a container name (e.g. swarms), I'd still like to see the *service* listed, along with the specific containers.

jesseduffield:
@pvande how about instead of having `lazydocker` in the status panel in the top left, we have the name of the current directory?

[...]

PR #88, opened 2019-07-03
Show Project Directory in Status Panel

pvande:
As per [this discussion](https://github.com/jesseduffield/lazydocker/issues/40#issuecomment-507624225), this puts the current working directory name into the Status Panel area.  As related changes, this PR also moves the "lazydocker" project name down to the information area, and renames the Status Panel to the "Project Panel" (which seemed a natural consequence).

### DIFF
--- a/pkg/gui/project_panel.go
+++ b/pkg/gui/project_panel.go
@@ -3,6 +3,7 @@ package gui
 import (
 	"bytes"
 	"fmt"
+	"path"
 	"strings"
 
 	"github.com/fatih/color"
@@ -30,9 +31,11 @@ func (gui *Gui) getProjectContextTitles() []string {
 func (gui *Gui) refreshProject() error {
 	v := gui.getProjectView()
 
+	dirName := path.Base(gui.Config.ProjectDir)
+
 	gui.g.Update(func(*gocui.Gui) error {
 		v.Clear()
-		fmt.Fprint(v, "lazydocker")
+		fmt.Fprint(v, dirName)
 		return nil
 	})
 
--- a/pkg/gui/project_panel.go
+++ b/pkg/gui/project_panel.go
@@ -31,11 +31,19 @@ func (gui *Gui) getProjectContextTitles() []string {
 func (gui *Gui) refreshProject() error {
 	v := gui.getProjectView()
 
-	dirName := path.Base(gui.Config.ProjectDir)
+	projectName := path.Base(gui.Config.ProjectDir)
+	if gui.DockerCommand.InDockerComposeProject {
+		for _, service := range gui.DockerCommand.Services {
+			if service.Container != nil {
+				projectName = service.Container.Details.Config.Labels["com.docker.compose.project"]
+				break
+			}
+		}
+	}
 
 	gui.g.Update(func(*gocui.Gui) error {
 		v.Clear()
-		fmt.Fprint(v, dirName)
+		fmt.Fprint(v, projectName)
 		return nil
 	})
 
--- a/pkg/gui/project_panel.go
+++ b/pkg/gui/project_panel.go
@@ -13,22 +13,22 @@ import (
 	"github.com/jesseduffield/yaml"
 )
 
-func (gui *Gui) getStatusContexts() []string {
+func (gui *Gui) getProjectContexts() []string {
 	if gui.DockerCommand.InDockerComposeProject {
-		return []string{"logs", "credits", "config"}
+		return []string{"logs", "config", "credits"}
 	}
 	return []string{"credits"}
 }
 
-func (gui *Gui) getStatusContextTitles() []string {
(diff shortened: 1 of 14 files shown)


pkg/gui/project_panel.go
```diff
@@ -176,28 +179,28 @@ func lazydockerTitle() string {
...
+	contexts := gui.getProjectContexts()
+	if gui.State.Panels.Project.ContextIndex >= len(contexts)-1 {
+		gui.State.Panels.Project.ContextIndex = 0
 	} else {
-		gui.State.Panels.Status.ContextIndex++
+		gui.State.Panels.Project.ContextIndex++
 	}
 
-	gui.handleStatusSelect(gui.g, v)
+	gui.handleProjectSelect(gui.g, v)
```
golangcibot: Error return value of `gui.handleProjectSelect` is not checked (from `errcheck`)

pkg/gui/project_panel.go
```diff
@@ -176,28 +179,28 @@ func lazydockerTitle() string {
...
+	contexts := gui.getProjectContexts()
+	if gui.State.Panels.Project.ContextIndex <= 0 {
+		gui.State.Panels.Project.ContextIndex = len(contexts) - 1
 	} else {
-		gui.State.Panels.Status.ContextIndex--
+		gui.State.Panels.Project.ContextIndex--
 	}
 
-	gui.handleStatusSelect(gui.g, v)
+	gui.handleProjectSelect(gui.g, v)
```
golangcibot: Error return value of `gui.handleProjectSelect` is not checked (from `errcheck`)

dawidd6:
Please also adjust this line:
https://github.com/jesseduffield/lazydocker/blob/572ff63e858150d9cbec6f1ed67ff6d04258c9c0/scripts/generate_cheatsheet.go#L29

pvande:
This would arguably be more correct if it preferred to use the "com.docker.compose.project" label...

``` sh
docker inspect --format='{{json .Config.Labels }}' $OBJECT_ID
```

``` json
{
"com.docker.compose.config-hash":"fa59a5b47221e5a6a9b4d4182709755a0079ae2b4b634f59c44108a6dd57d081",
"com.docker.compose.container-number":"1",
"com.docker.compose.oneoff":"False",
"com.docker.compose.project":"project-name",
"com.docker.compose.service":"anonymous-service",
"com.docker.compose.version":"1.23.2"
}
```

I don't know how often anyone actually [customizes the project name](https://docs.docker.com/compose/reference/overview/#use--p-to-specify-a-project-name), though...

jesseduffield:
@pvande coincidentally, my work does haha. I'm not sure what the best approach is here. Is there a way to check if the project name hasn't been custom-set, and if so just revert to using the initial approach?

pvande:
@jesseduffield Awesome!  As near as I can tell, the only way to get the project name back out of docker-compose is through the container labels.  That's theoretically easy enough, though it introduces a strange cross-panel dependency (the Status/Project Panel now needs the list of Services).

I'll see if I can get a PoC together quickly.

pkg/gui/project_panel.go
```diff
@@ -13,33 +14,43 @@ import (
 	"github.com/jesseduffield/yaml"
 )
 
-func (gui *Gui) getStatusContexts() []string {
+func (gui *Gui) getProjectContexts() []string {
 	if gui.DockerCommand.InDockerComposeProject {
-		return []string{"logs", "credits", "config"}
+		return []string{"logs", "config", "credits"}
```
jesseduffield: I'm cool with this

jesseduffield:
LGTM. One last thing though, given that we will likely have a bit of free horizontal space in the project panel, I wonder if we should add in brackets the docker-compose command from the config. That way, if people start lazydocker like `lazydocker -f docker-compose.test.yml` we can say
```
your-project-name (docker-compose -f docker-compose.test.yml)
```

That way they have even more context about what they're seeing. What are your thoughts on that?

pvande:
@jesseduffield I'm not opposed to surfacing the `docker-compose` command, but I'm not myself convinced that after the project name in the project panel is the best place for it.

While we will have some space after the project name, even the simple example you gave is 62 characters long.  At my desk, on my 4K monitor, that would probably fit just fine, but when I'm out on my laptop, my entire terminal width is commonly between 110 and 230 characters.  That doesn't really leave enough space to show any meaningful amount of that command.

It might be better to put that information on a second line in the project panel (space is still an issue, but less so), or in a tab under the project panel (perhaps fleshing the `config` tab out with more than just the literal config text).

jesseduffield:
I agree. We can have another look at that after this is merged. As for this PR, it LGTM. Anything else you needed to do before we merge it?

pvande:
Anything further I might contribute fits just as easily in a new PR.  Merge away!  🙂

jesseduffield:
thanks again :)

PR #96, opened 2019-07-04
CI snapcraft installation

dawidd6:


dawidd6:
Should work, tested it in Docker on image that is used by CI.
```sh
# testing procedure in Docker container based on circleci/golang:1.12 image
# while being in lazydocker source code dir
./.circleci/install_snapcraft.sh
curl -sL https://git.io/goreleaser | bash -s -- --skip-publish --snapshot --rm-dist --skip-sign
```

dawidd6:
Hmm. We can always make our own image and push it to Docker Hub with everything needed already in it. Or try to contribute to goreleaser directly by adding snapcraft to their Dockerfile.

But for now this solution may be enough.

PR #97, opened 2019-07-04
Added rounding to timing (fixes #94)

### LINKED ISSUE #94: Round seconds in stats
So many digits for seconds in stats, i think it will look better without that. Maybe round to 2 digits?
decimal

mjarkk:
This fixes #94  
Now it shows: `1h27m13s` in staid of `1h27m13.246503645s`


jesseduffield:
LGTM

PR #98, opened 2019-07-05
Add a limit to the c.StatsHistory

### LINKED ISSUE #76: Keeping stats view open for a while produces `index out of range` error
**Describe the bug**
Keeping the stats window open for a given container for long periods of time eventually crashes with `panic: runtime error: index out of range`.

**To Reproduce**
1. selecting a running container
2. select stats view
3. wait for a while (I've seen it happen in a couple of hours)
4. see error

**Expected behaviour**
Stats view to not crash ever

**Screenshots**
Screenshot from 2019-07-03 05-27-20

 - OS: Debian Linux 9.9
 - Lazydocker `commit=50a6eb64ece25bc85c95c7223f6100bdbae6c69c, build date=2019-06-30T11:09:18Z, build source=binaryRelease, version=0.2.4, os=linux, arch=amd64`

mjarkk:
While trying to fix #76 i found that the stats didn't have a limit.  
This might be why lazydocker seems to crash over a long time but from the time in #76 screenshots (17498s what is equal to the amount of array items) this seems wired because a slice can have more than that amount of items  

Beside that i don't think that it's a good idea looping every second over an array with 10000+ items.

### DIFF
--- a/pkg/commands/docker.go
+++ b/pkg/commands/docker.go
@@ -179,12 +179,7 @@ func (c *DockerCommand) createClientStatMonitor(container *Container) {
 		}
 
 		c.ContainerMutex.Lock()
-		toSlice := len(container.StatHistory) - 50
-		if toSlice < 0 {
-			toSlice = 0
-		}
-		container.StatHistory = append(container.StatHistory[toSlice:], recordedStats)
-
+		container.StatHistory = append(container.StatHistory, recordedStats)
 		container.EraseOldHistory()
 		c.ContainerMutex.Unlock()
 	}
--- a/pkg/commands/docker.go
+++ b/pkg/commands/docker.go
@@ -179,7 +179,12 @@ func (c *DockerCommand) createClientStatMonitor(container *Container) {
 		}
 
 		c.ContainerMutex.Lock()
-		container.StatHistory = append(container.StatHistory, recordedStats)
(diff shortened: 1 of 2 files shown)


jesseduffield:
Oops! The actual issue here is that there's no default max duration set:
```
stats:
  maxDuration: 5m
```

jesseduffield:
this should do the trick:
```
diff --git a/pkg/config/app_config.go b/pkg/config/app_config.go
index 3976ad9..4b87098 100644
--- a/pkg/config/app_config.go
+++ b/pkg/config/app_config.go
@@ -289,6 +289,11 @@ type CustomCommand struct {
 // the boolean zero value and this will be ignored when parsing the user's
 // config
 func GetDefaultConfig() UserConfig {
+	duration, err := time.ParseDuration("5m")
+	if err != nil {
+		panic(err)
+	}
+
 	return UserConfig{
 		Gui: GuiConfig{
 			ScrollHeight:      2,
@@ -336,6 +341,7 @@ func GetDefaultConfig() UserConfig {
 			Method: "never",
 		},
 		Stats: StatsConfig{
+			MaxDuration: duration,
 			Graphs: []GraphConfig{
 				{
 					Caption:  "CPU (%)",

```

mjarkk:
Thanks will change that later today

jesseduffield:
LGTM, mergin

Issue #64, opened 2019-07-02
[Question] What is the minimum supported docker version?


warmchang:
**Describe the bug**
lazydocker return an error.

**To Reproduce**
Run lazydocker with docker v1.12.6 on CentOS 7.

**Expected behaviour**
Run correctly.

**Screenshots**
![image](https://user-images.githubusercontent.com/13084624/60505738-e9358f00-9cf6-11e9-9e7a-216b29ea692e.png)

**Desktop (please complete the following information):**
 - OS:  CentOS 7
 - Lazydocker Version v0.3
 - The last commit id if you built project from sources (run : ```git-rev parse HEAD```) N/A

**Additional context**
```
[root@paas-controller-172-160-20-121:/home/ubuntu]$ docker version
Client:
 Version:      1.12.6
 API version:  1.24
 Go version:   go1.7.5
 Git commit:   2e3ea0a2a31be47d1baacde97b3a68e3ce32a8ab
 Built:        2019-06-06 20:56:02
 OS/Arch:      linux/amd64

Server:
 Version:      1.12.6
 API version:  1.24
 Go version:   go1.7.5
 Git commit:   2e3ea0a2a31be47d1baacde97b3a68e3ce32a8ab
 Built:        2019-06-06 20:56:02
 OS/Arch:      linux/amd64
[root@paas-controller-172-160-20-121:/home/ubuntu]$
[root@paas-controller-172-160-20-121:/home/ubuntu]$ ./lazydocker
2019/07/03 01:30:07 An error occurred! Please create an issue at https://github.com/jesseduffield/lazydocker/issues

*errors.errorString Error response from daemon: client is newer than server (client API version: 1.25, server API version: 1.24)
/go/src/github.com/jesseduffield/lazydocker/main.go:71 (0x8234e8)
/usr/local/go/src/runtime/proc.go:200 (0x42d77c)
/usr/local/go/src/runtime/asm_amd64.s:1337 (0x458541)
[root@paas-controller-172-160-20-121:/home/ubuntu]$

```

dawidd6:
Perhaps we could change:
```go
cli, err := client.NewEnvClient()
```
to
```go
cli, err := client.NewClientWithOpts(client.WithVersion("1.24"))
```

warmchang:
@dawidd6 thank you for replay.

Do this change:
```
--- a/pkg/commands/docker.go
+++ b/pkg/commands/docker.go
[...]

PR #102, opened 2019-07-05
Docker SDK update

dawidd6:
This should theoretically close: #64 

I updated the Docker SDK used in lazydocker from `v1.13.1` (which was released in Feb 8, 2017) to `v0.7.3-0.20190307005417-54dddadc7d5d`. Also specified the minimum Docker API version supported to `1.20`.

Why not a normal tag, but a commit?
Well, cause when one would have a look at https://github.com/moby/moby then one would see that the latest tag was pushed on Jun 28, 2017.
So, this PR is simply bumping Docker dependency to `54dddadc7d5d` commit.

Why this commit specifically?
I just use it in my project and it works well, so thought it might be okay to use it here as well.

---

**Note for myself**: update README requirements section, when this is merged.


jesseduffield:
LGTM. I'm guessing all those extra lines are just a reflection of how that package has grown over time? Also have you done a quick test of lazydocker's basic functionality to check that nothing was broken by this change?

dawidd6:
That's right, they must have changed something.

Yes, I did some testing, seems to be ok.

jesseduffield:
sounds good :)

dawidd6:
So, I roughly tested it with Docker in Docker image, cause I couldn't find a better way to install older version of Docker and it seems to work, but I think we can wait for @warmchang to test it on his machine.

warmchang:
FYI: https://github.com/jesseduffield/lazydocker/issues/64#issuecomment-509049007

jesseduffield:
Looks like we can merge this then, you just need to resolve conflicts @dawidd6

dawidd6:
Yep. Will do later today.

PR #107, opened 2019-07-06
Support bulk commands in each of the panels

jesseduffield:
This PR:
1) Rolls out the custom commands functionality to each of the side panels (except the top side panel)
2) Adds some configurable bulk commands like `docker-compose up -d` and `docker container prune` for the side panels.

### DIFF
--- a/pkg/gui/containers_panel.go
+++ b/pkg/gui/containers_panel.go
@@ -11,6 +11,7 @@ import (
 	"github.com/go-errors/errors"
 	"github.com/jesseduffield/gocui"
 	"github.com/jesseduffield/lazydocker/pkg/commands"
+	"github.com/jesseduffield/lazydocker/pkg/config"
 	"github.com/jesseduffield/lazydocker/pkg/utils"
 )
 
@@ -476,8 +477,8 @@ func (gui *Gui) handleContainerAttach(g *gocui.Gui, v *gocui.View) error {
 	return gui.Errors.ErrSubProcess
 }
 
-func (gui *Gui) handlePruneContainers(g *gocui.Gui, v *gocui.View) error {
-	return gui.createConfirmationPanel(gui.g, v, gui.Tr.Confirm, gui.Tr.ConfirmPruneContainers, func(g *gocui.Gui, v *gocui.View) error {
+func (gui *Gui) handlePruneContainers() error {
+	return gui.createConfirmationPanel(gui.g, gui.getContainersView(), gui.Tr.Confirm, gui.Tr.ConfirmPruneContainers, func(g *gocui.Gui, v *gocui.View) error {
 		return gui.WithWaitingStatus(gui.Tr.PruningStatus, func() error {
 			err := gui.DockerCommand.PruneContainers()
 			if err != nil {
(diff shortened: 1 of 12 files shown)


pkg/gui/images_panel.go
```diff
@@ -259,3 +259,18 @@ func (gui *Gui) handlePruneImages(g *gocui.Gui, v *gocui.View) error {
 		})
 	}, nil)
 }
+
+func (gui *Gui) handleImagesCustomCommand(g *gocui.Gui, v *gocui.View) error {
```
golangcibot: U1000: func `(*Gui).handleImagesCustomCommand` is unused (from `unused`)

pkg/gui/volumes_panel.go
```diff
@@ -250,3 +250,18 @@ func (gui *Gui) handlePruneVolumes(g *gocui.Gui, v *gocui.View) error {
 		})
 	}, nil)
 }
+
+func (gui *Gui) handleVolumesCustomCommand(g *gocui.Gui, v *gocui.View) error {
```
golangcibot: U1000: func `(*Gui).handleVolumesCustomCommand` is unused (from `unused`)

mcintyre94:
Nice one! I really like the use of a separate key to open bulk shortcuts, looks like it'll work much better than combining with the x shortcuts since it's being applied to all panels. Also good shout to expand to all side panels :)

> Does this look right?

Looks really good to me - definitely got the ones I'd be looking for there.

> Should we now remove the hardcoded shift+d keybinding which was for pruning all containers, given that you'll now be able to do it through the menu?

Probably, using the bulk menu seems clearer IMO.

> Did we have plans to remove all CLI usage at some point? If so, this is adding tech debt. Although it's worth considering whether configurability is more important than whatever we hoped to get out of moving away from the CLI commands

I wasn't aware of that, do you know where it was discussed? Agree this adds tech debt if that was a goal, not sure how else we'd do it though. Is there a go lib for docker-compose like there is for docker (assuming that's what docker CLI would be replaced with)?

> Should we have a waiting status (the thing you see at the bottom left corner while waiting) for each of these commands? Currently it just says 'running bulk command...'

Seems pretty minor, but maybe we could customise it per panel? 'running bulk container command' / 'running bulk volume command' etc.?

pkg/gui/containers_panel.go
```diff
@@ -517,3 +517,29 @@ func (gui *Gui) handleContainersCustomCommand(g *gocui.Gui, v *gocui.View) error
 
 	return gui.createCustomCommandMenu(customCommands, commandObject)
 }
+
+func (gui *Gui) handleStopContainers(g *gocui.Gui, v *gocui.View) error {
+	return gui.createConfirmationPanel(gui.g, v, gui.Tr.Confirm, gui.Tr.ConfirmStopContainers, func(g *gocui.Gui, v *gocui.View) error {
+		return gui.WithWaitingStatus(gui.Tr.StoppingStatus, func() error {
+
+			for _, container := range gui.DockerCommand.Containers {
+				container.Stop()
```
golangcibot: Error return value of `container.Stop` is not checked (from `errcheck`)

pkg/gui/containers_panel.go
```diff
@@ -517,3 +517,29 @@ func (gui *Gui) handleContainersCustomCommand(g *gocui.Gui, v *gocui.View) error
...
+		})
+	}, nil)
+}
+
+func (gui *Gui) handleRemoveContainers(g *gocui.Gui, v *gocui.View) error {
+	return gui.createConfirmationPanel(gui.g, v, gui.Tr.Confirm, gui.Tr.ConfirmRemoveContainers, func(g *gocui.Gui, v *gocui.View) error {
+		return gui.WithWaitingStatus(gui.Tr.RemovingStatus, func() error {
+
+			for _, container := range gui.DockerCommand.Containers {
+				container.Remove(types.ContainerRemoveOptions{Force: true})
```
golangcibot: Error return value of `container.Remove` is not checked (from `errcheck`)

jesseduffield:
Alright I've made some changes. Now we're not using the docker CLI directly for anything, we're using the docker SDK.

I've added a new key on our custom command struct called InternalFunction which is the name of a whitelisted internal function e.g. 'pruneContainers'. This is so that we can have custom commands that point to internal functions sitting alongside custom commands that have their own CLI commands.

I am a little concerned about the fact that I'm introducing a concept that's not only not type safe, but  is spread across a few files. The mapping of whitelisted functions is in custom_commands.go, but referenced in app_config.go, and the actual functions themselves are in the corresponding _panels file. I can't move the mapping into app_config.go because that file has no concept of the gui struct.

An alternative would be to have some stock standard custom commands that are not configurable, and those are for the internal functions. Then the remaining custom commands are configurable. I'm leaning towards this approach, but it's going to be a bit messier than what we have now.

Any thoughts @mcintyre94 ?

mcintyre94:
Unfortunately I'm not very well positioned to do Go code review! I can see what you're refering to though, it looks like it could get a bit tricky to maintain. I wonder if it'd make sense to add a struct for the valid keys of that map, which could (I think?) be available to `app_config.go` as well as `custom_commands.go`? Might just be adding another layer of indirection, but it'd at least be really obvious if something was added to `custom_commands.go` that wasn't in that keys map, and at least saves any typos referencing them in `app_config.go`. 

Or is that what you meant by having some stock commands that aren't configurable? I think I would lean toward something a bit safer but not sure what the best way of achieving that would be.

jesseduffield:
Yeah it seems like the right way to do this is to have some bulk commands that are there by default, which aren't configurable and just reference internal functions, and then allow the user to add some extra bulk commands on top which can't reference internal functions. Then we don't need to worry about type safety or keeping things synced up or anything. I'll try to make that happen this week.

jesseduffield:
does anybody want to have a look at this PR before I merge? @dawidd6 @mjarkk? @glvr182

dawidd6:
It would be also a good idea to update README if you bumped API version.

Issue #110, opened 2019-07-06
Shortcut for hiding exited containers


mjarkk:
**Is your feature request related to a problem? Please describe.**
When i'm working on and testing a Dockerfile lazydocker shows all the stopped containers in the container overview.  
The problem with this is that i need to scroll far down to see any of the real running containers.  
I can't see any reason for showing the stopped containers beside letting me know that I've lots of docker junk and need to press `D`.  
![image](https://user-images.githubusercontent.com/15320763/60759573-a130a800-a027-11e9-854f-c503192e0dfc.png)

**Describe the solution you'd like**
A shortcut for showing or hiding the stopped containers.  
Because `docker ps` by default doesn't show stopped containers it would be nice if lazydocker would do the same.  

**Describe alternatives you've considered**
We cloud also make this a config setting.

mjarkk:
This is now fixed in v0.5.5

PR #111, opened 2019-07-06
Added `e` keybinding for toggling visibility stopped containers

mjarkk:
This is a suggested fix for #110  
@jesseduffield I've added an translation struct item to the `TranslationSet` but i'm not sure what to do next, shall i add it manually add it to every translation file and keybindings readme file or can that be dune semi automated?


pkg/commands/docker.go
```diff
@@ -29,6 +29,7 @@ type DockerCommand struct {
 	Config                 *config.AppConfig
 	Client                 *client.Client
 	InDockerComposeProject bool
+	ShowExited             bool
```
jesseduffield: I would call this HideExited so that it defaults to false. There are two reasons for this:
1) I think people generally want to see when they've got exited containers
2) Exited containers still take up space on the disk. If somebody has a lot of exited containers, they should probably to a `docker container prune`

pkg/gui/containers_panel.go
```diff
@@ -277,7 +277,17 @@ func (gui *Gui) refreshContainersAndServices() error {
 	gui.g.Update(func(g *gocui.Gui) error {
 		containersView.Clear()
 		isFocused := gui.g.CurrentView().Name() == "containers"
-		list, err := utils.RenderList(gui.DockerCommand.DisplayContainers, utils.IsFocused(isFocused))
+		toRender := gui.DockerCommand.DisplayContainers
```
jesseduffield: filtering the containers here won't work because we use the currently selected line index as the index on DisplayContainers so if you're hiding a heap of exited containers and now the first container in the list actually has index 10 in DisplayContainers, you won't see it's logs/graph. I've just tested to verify this.

We should instead do this filtering in gui.RefreshContainersAndServices() when we create DisplayContainers.

pkg/commands/docker.go
mjarkk: Agree with you,  
I'll change it then to be default: `true`

mjarkk:
Jup agree with your point about showing the exited containers,  
I'll change it then to shows them by default.

mjarkk:
@jesseduffield I've changed and dune what you've suggested

pkg/i18n/english.go
```diff
@@ -115,6 +116,7 @@ func englishSet() TranslationSet {
 		EditConfig:         "edit lazydocker config",
 		Cancel:             "cancel",
 		Remove:             "remove",
+		HideStopped:        "Hide/Show stopped container",
```
jesseduffield: I think the plural makes more sense here
```suggestion
		HideStopped:        "Hide/Show stopped containers",
```

jesseduffield:
@mjarkk there is an issue in using the 'h' keybinding because that's used to cycle through the side panels. We will need to use another keybinding :P

mjarkk:
Ow i didn't know that :(,  
Do you have another suggested key?

jesseduffield:
I can't really think of a key that's appropriate. maybe go for 'I' or 'U'? I don't think they're being used

mjarkk:
I've changed it to `e`

jesseduffield:
good enough for me!

jesseduffield:
btw you won't need to update the cheatsheet manually anymore, it'll be done automatically with each new release :)

PR #112, opened 2019-07-06
Cleaned up some code from lazygit

mjarkk:


coverage.txt
```diff
@@ -1,87 +1,5 @@
 
 mode: atomic
-github.com/jesseduffield/lazydocker/pkg/commands/container.go:247.64,249.2 1 0
```
jesseduffield: is this file supposed to be committed?

jesseduffield:
LGTM

jesseduffield:
although I'm still interested in knowing: should the coverage.txt file be committed @mjarkk ?

mjarkk:
@jesseduffield I think it's a better idea to ignore the coverage file by default for cleaner commits.

coverage.txt
mjarkk: No but because i ran the ./test.sh it gave me this what i accidentally commited because of a `git add .`

PR #114, opened 2019-07-07
update docs when merging to master

jesseduffield:
Now we automatically generate the new keybinding cheatsheets upon merging a PR into master


jesseduffield:
actually this makes more sense to only be done upon releasing, So I've moved it to there

PR #116, opened 2019-07-07
Attach to container via SDK

dawidd6:
### Showcase (ec5f96a1eba8e793025636823d51d268eaf286ae)
[asciicast](https://asciinema.org/a/ExadsbEj2pDDac75nxn6HWkII)

### Current status
\+ detaching from container works after pressing standard Docker key sequence (ctrl-p ctrl-q)
\+ panels are hiding when attaching
\+ panels are showing back when detaching
\+ nicely returns to lazydocker when container exits
\+ text is always white and not bold


pkg/commands/container.go
```diff
@@ -341,18 +346,80 @@ func (c *Container) Restart() error {
...
+
+	if c.Details.Config.OpenStdin {
+		fd := int(os.Stdin.Fd())
+
+		if terminal.IsTerminal(fd) {
+			oldState, err := terminal.MakeRaw(fd)
+			if err != nil {
+				return err
+			}
+			defer terminal.Restore(fd, oldState)
```
golangcibot: Error return value of `terminal.Restore` is not checked (from `errcheck`)

pkg/commands/container.go
```diff
@@ -341,18 +346,80 @@ func (c *Container) Restart() error {
...
+
+			go c.resizeIfChanged(fd)
+			go io.Copy(hijack.Conn, os.Stdin)
```
golangcibot: Error return value of `io.Copy` is not checked (from `errcheck`)

pkg/commands/container.go
```diff
@@ -341,18 +346,80 @@ func (c *Container) Restart() error {
...
+	return nil
+}
+
+func (c *Container) resizeIfChanged(fd int) {
+	channel := make(chan os.Signal)
+	signal.Notify(channel, syscall.SIGWINCH)
+
+	for {
+		<-channel
+		c.Resize(fd)
```
golangcibot: Error return value of `c.Resize` is not checked (from `errcheck`)

pkg/gui/containers_panel.go
```diff
@@ -467,12 +467,13 @@ func (gui *Gui) handleContainerAttach(g *gocui.Gui, v *gocui.View) error {
...
 	}
 
-	c, err := container.Attach()
+	err = container.Attach()
 	if err != nil {
 		return gui.createErrorPanel(gui.g, err.Error())
 	}
 
-	gui.SubProcess = c
+	gui.RunWithSubprocesses()
```
golangcibot: Error return value of `gui.RunWithSubprocesses` is not checked (from `errcheck`)

pkg/commands/container.go
```diff
@@ -341,18 +346,80 @@ func (c *Container) Restart() error {
...
+	}
+
+	hijack.Close()
+
+	return nil
+}
+
+func (c *Container) resizeIfChanged(fd int) {
+	channel := make(chan os.Signal)
+	signal.Notify(channel, syscall.SIGWINCH)
```
golangcibot: SA1017: the channel used with signal.Notify should be buffered (from `staticcheck`)

pkg/commands/container.go
```diff
@@ -341,22 +349,102 @@ func (c *Container) Restart() error {
...
+			// shamelessly taken from docker/cli
+			escapeKeys := []byte{16, 17}
+			// Stop reading if escape sequence had been entered
+			input := term.NewEscapeProxy(os.Stdin, escapeKeys)
+			_, _ = io.Copy(hijack.Conn, input)
+			channel <- syscall.SIGINT
+		}()
+	}
+
+	for {
```
golangcibot: S1000: should use for range instead of for { select {} } (from `gosimple`)

dawidd6:
I'm marking this PR "Ready to review", just cause I can see then if there are any merge conflicts.

pkg/commands/container.go
```diff
@@ -341,22 +348,101 @@ func (c *Container) Restart() error {
...
+			if err != nil {
+				return err
+			}
+		} else {
+			err = term.RestoreTerminal(fd, oldState)
+			if err != nil {
+				return err
+			}
+
+			hijack.Close()
```
jesseduffield: this should be `defer hijack.Close()` right after we initialize it so that if we ever return early due to an error the session will be closed

pkg/commands/container.go
```diff
@@ -341,22 +348,101 @@ func (c *Container) Restart() error {
...
+
+	channel := make(chan os.Signal, 1)
+	signal.Notify(channel, syscall.SIGWINCH)
+	// initial resize
+	channel <- syscall.SIGWINCH
+
+	// read output from container
+	go func() {
+		output := os.Stdout
+		_, _ = io.Copy(output, hijack.Conn)
```
jesseduffield: I did some digging, and it turns out you can add the following to clear the screen and unhide the cursor:
```
		// clear the screen and unhide the cursor
		_, _ = io.Copy(output, strings.NewReader("\033[2J\033[?25h"))
```
This fixes the issue with the panels showing up in the background

pkg/commands/container.go
```diff
@@ -341,22 +348,101 @@ func (c *Container) Restart() error {
...
+		channel <- syscall.SIGINT
+	}()
+
+	// send input to container
+	if c.Details.Config.OpenStdin {
+		go func() {
+			// the default escape key sequence: ctrl-p, ctrl-q
+			//
+			// shamelessly taken from docker/cli
+			escapeKeys := []byte{16, 17}
```
jesseduffield: would be cool if this was customisable, but we can leave it for now

pkg/commands/container.go
```diff
@@ -341,22 +348,101 @@ func (c *Container) Restart() error {
...
+
+	for {
+		sig := <-channel
+		if sig == syscall.SIGWINCH {
+			err := c.Resize(fd)
+			if err != nil {
+				return err
+			}
+		} else {
+			err = term.RestoreTerminal(fd, oldState)
```
jesseduffield: I wonder if we should use this for our subprocesses. Currently we close and re-initialize the gui library, but it looks like we don't need to be.

jesseduffield: (that can be left to another PR though)

pkg/commands/container.go
dawidd6: Nice find. Works just right :+1:

pkg/commands/container.go
dawidd6: Agreed, but for this to be customizable fully, I think there is a need for https://github.com/jesseduffield/lazydocker/pull/117 to be merged first.

dawidd6:
> Only issues from my end is that it doesn't show you the actual current position of the cursor until you hit return, but I'm not sure how to get around that.

Yes, I have the same "issue". I tinkered with this today and found out that resizing terminal just a tiny bit makes the prompt appear with cursor in proper place like it should be. But I couldn't achieve this effect in code without actual manual intervention with mouse.
 
> Another things is that every time I switch, A bell rings in my terminal (iterm2, OSX). Does that happen to you as well? I imagine it would get quite annoying

I have terminal (gnome-terminal, Ubuntu) bell switched off, but after switching it on I can't hear anything, strange.

mjarkk:
> I have terminal (gnome-terminal, Ubuntu) bell switched off, but after switching it on I can't hear anything, strange.  
You could try the [elementry os terminal](https://github.com/elementary/terminal), that one also rings a bell for every small thing.

dawidd6:
> Only issues from my end is that it doesn't show you the actual current position of the cursor until you hit return, but I'm not sure how to get around that.

Okay, I think I resolved this in latest commit. Once againg `docker/cli` repo was helpful.

@mjarkk Don't want to sound rude or something, but I'm too lazy to set up an elementary VM :). Really I have no idea what would cause that in code.

pkg/commands/container.go
```diff
@@ -436,6 +450,11 @@ func (c *Container) Resize(fd uintptr) error {
 		Width:  uint(size.Width),
 	}
 
+	if fool {
```
jesseduffield: good idea, but unfortunately it doesn't work on iterm2 for me :( I still have to press enter to show the prompt. hmm, this is tricky

dawidd6: Does the resizing even work at all? Try to write some characters, then resize terminal and see if lines wrap.

jesseduffield: if you do this:
```
		if sig == syscall.SIGWINCH {
			err := c.Resize(fd, false)
			if err != nil {
				return err
			}
			moveCursorRight := "\033[1C"
			hijack.Conn.Write([]byte(moveCursorRight))
```

It works for me. Only issue is, we don't actually want to send a command that interacts with the underlying process. If we could find an ansi escape code that causes a re-render, but doesn't actually change anything, that would be the proper solution!

jesseduffield: (didn't see your comment above). Yeah resizing works fine for me

jesseduffield:
thinking out loud about the bell. I wonder if we could just filter out any bell characters from the output (i.e. "\a") for the first second. Crude solution, but would get the job done

dawidd6:
Not working on this now.

PR #117, opened 2019-07-07
[WIP] Configurable keybinds

### LINKED ISSUE #48: Configurable keybindings
Default keybindings are hard coded. Problem is, not every keyboard has `[` and `]` keys.

Make the keybindings configurable.

Move the `previous/next tab` keybindings from `[` and `]` to something else.

glvr182:
fixes #48 
Signed-off-by: Glenn Vriesman <glenn.vriesman@gmail.com>


pkg/config/app_config.go
```diff
@@ -250,6 +250,72 @@ type StatsConfig struct {
...
+	ScrollUpMain    []string
+	ScrollDownMain  []string
+	AutoScrollMain  []string
+	ShowOptionsMenu []string
+	CustomCommand   []string
+	Project         struct {
+		EditConfig      []string
+		OpenConfig      []string
+		PreviousContext []string
+		NextContext     []string
```
jesseduffield: I think cross-cutting keybindings like NextContext and PreviousContext should sit at the top level rather than under each panel that uses them. That way the user only needs to override the key in one place.

glvr182: Agreed, as I copied everything from the keybindings and i found a lot of repeating configs, will look more into that

glvr182:
@jesseduffield I'm in a bit of a pickle, keybindings.ParseAll() returns both the an object(containing the key, the modifier and the tokens), and an error. How should i go about setting the keybindings in keybinding.go? I was thinking about a for loop, but that seems like a lot of work for a not so pretty solution. I'm now thinking of a helper function that panics if it contains an error and having a recovery function to catch that, but even with that i'd have to refactor a lot

glvr182:
The keybinding struct (now having []interface{} instead of interface{})
```
type Binding struct {
  ViewName    string
  Handler     func(*gocui.Gui, *gocui.View) error
  Key         []interface{} // FIXME: find out how to get `gocui.Key | rune`
  Modifier    gocui.Modifier
  Description string
}
```

The recovery
```
defer func() {
  if r := recover(); r != nil {
    fmt.Println("Recovered in f", r)
  }
}()
```
The keybinding def
```
{
  ViewName: "",
  Key:      filterParser(keybinding.ParseAll("ctrl+A, ctrl+B, CTRL+ALT+C")),
  Modifier: gocui.ModNone,
  Handler:  gui.quit,
},
```
The helper function
```
func filterParser(binds []keybinding.Key, err error) []gocui.Key {
  var keys []gocui.Key
  if err != nil {
    panic(err)
  }
  for _, bind := range binds {
    keys = append(keys, bind.Value)
  }
  return keys
}
```

jesseduffield:
@glvr182 I think we should put the default KeyConfig on the UserConfig struct in app_config.go. We can then have our filterParser method in that file and it can handle all of our defaults on startup.

In UserConfig:
```
Keybindings KeyConfig `yaml:"keybindings,omitempty"`
```

In GetDefaultConfig
```
		Keybindings: KeyConfig{
			Quit: []string{"q"},
                         ...
		},
```

Then in keybindings.go
```
		{
			ViewName: "",
			Keys:     parsedKeybindings(gui.Config.UserConfig.Keybindings.Quit),
			Modifier: gocui.ModNone,
			Handler:  gui.quit,
		},
```

and then in the same file:
```
func parsedKeybindings(binds []string) []gocui.Key {
  var keys []gocui.Key
  for _, bind := range binds {
    keys = append(keys, keybindings.MustParse(bind)) // MustParse panics on failure
  }
  return keys
}
```

you'll need each keybinding to support an array of keys, so then you'll be doing this in `keybindings()`:
```
	for _, binding := range bindings {
		for _, key := range binding.keys {
			if err := g.SetKeybinding(binding.ViewName, key, binding.Modifier, binding.Handler); err != nil {
				return err
			}
		}
	}
```

I think it's worth adding a MustParse() function to the keybindings library, which panics on failure, so that we don't need to worry about the error that could be returned. As you say we can recover from it.

Hopefully the fact we now need to use an []interface{} isn't too bad, though I have had bad experiences with interface{} arrays in the past.

What are your thoughts?

glvr182:
I think having a must parse function in the lib is a good idea, I shall make that function tomorrow, (or if you have time be my guest).

I'm not the biggest fan of interfaces either, i prefer having an actual type, but maybe we find a way to resolve that along the road. (might try that in this pr).

I think with the ```MustParse``` function I'll be able to finish this soon

glvr182:
Just finished the mustParse functions, will use them now

glvr182:
@jesseduffield I'm again in a bit of a pickle, keybinding library uses awesome-gocui/gocui, lazydocker uses jesseduffield/gocui, what do you recon would be the best

jesseduffield:
@glvr182 that is a good question. I would like for the two gocuis to be one-and-the-same, but I haven't really put the effort in to port my stuff over to awesome-gocui/gocui when I add something to my own.

What are your thoughts on defining the gocui keys in the keybindings package and casting them in the consuming package? I'm assuming they are all integers under the hood? Otherwise, if we think that it's going to be too much work to get the two gocuis synced up, I could fork the keybindings package and get it to point at my own gocui.

glvr182:
They are as far as i know [uint16](https://github.com/awesome-gocui/termbox-go/blob/HEAD/api_common.go), I don't really like the idea of casting, for obvious reasons, but as you said we would have to update awesome-gocui/gocui a lot. Maybe we could focus on that, as others might have use for it aswell.

jesseduffield:
agreed

glvr182:
I'm settings the defaults right now, it's taking longer than i hoped, thanks for the patience

glvr182:
Rebased from master

glvr182:
@jesseduffield might it be possible to lock the app_config.go and keybindings files for the time being? Just untill the pr is done

jesseduffield:
sure, how long do you think you'll need?

glvr182:
Well once i set all the defaults, i can test it. So how about this weekend?

glvr182:
Apart from the errors, i think you can have a look at the keybinding configs @jesseduffield

glvr182:
Seems to be an issue with parsing characters to keys, will have to do more investigation
Might end up merging keybindings into gocui

glvr182:
Hit a bug while setting up the parser, hope to get stuff working tomorrow

dawidd6:
@glvr182 how is it going? Maybe need some halp?

glvr182:
Yea well I back then hit a bug that had to do with some casting stuff. Haven't been able to fix that, but i was working on merging jesse's gocui with the awesome-gocui/gocui one to fix it. I'll post some errors tonight.

coolCucumber-cat:
It still doesn't work, even though it was merged into main. What's going on?

PR #127, opened 2019-07-12
Allow user setting the pull interval

### LINKED ISSUE #115: High CPU usage by Docker for Mac when lazydocker is running idle
**Describe the bug**
When running `lazydocker` on my Mac (latest version), while no containers are running, Docker for Mac uses ~25-30% cpu.

**To Reproduce**
1. Start `lazydocker`
2. Open `Activity Monitor` in CPU tab
3. Find process `com.docker.hyperkit`
4. See CPU usage

**Expected behaviour**
CPU usage low (without `lazydocker` it's ~2%). Maybe small impact.

**Screenshots**
NA

 - OS: Mac OS (latest as of July 7th 2019).
 - Lazydocker Version `Version: commit=a536c09dab90fc9a4f4931285466d6126c8bc121, build date=2019-07-05T10:59:26Z, build source=binaryRelease, version=0.5.4, os=darwin, arch=amd64`
 - The last commit id if you built project from sources (run : ```git rev-parse HEAD```)

**Additional context**
I assume the load is due to continuous querying of the Docker API. Might be a good idea to allow setting the polling interval so I can choose to config a lower polling rate and reduce the load on Docker.

mjarkk:
This fixes #115,  
With this you can add:  
```yml
update:
  refreshProjectTime: 100ms
  refreshContainersAndServicesTime: 100ms
  refreshVolumesTime: 100ms
```
To the lazydocker config to change the pull timing.

### DIFF
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -53,9 +53,7 @@ oS:
   openCommand: open {{filename}}
   openLinkCommand: open {{link}}
 update:
-  refreshProjectTime: 100ms
-  refreshContainersAndServicesTime: 100ms
-  refreshVolumesTime: 100ms
+  dockerRefreshInterval: 100ms
   method: never
 stats:
   graphs:
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -13,7 +13,7 @@ Changes to the user config will only take place after closing and re-opening laz
 
 ## Default:
 
-```
+```yml
 gui:
   scrollHeight: 2
   theme:
@@ -53,6 +53,9 @@ oS:
   openCommand: open {{filename}}
   openLinkCommand: open {{link}}
 update:
+  refreshProjectTime: 100ms
+  refreshContainersAndServicesTime: 100ms
+  refreshVolumesTime: 100ms
   method: never
 stats:
   graphs:
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -54,7 +54,6 @@ oS:
   openLinkCommand: open {{link}}
 update:
   dockerRefreshInterval: 100ms
-  method: never
 stats:
   graphs:
   - caption: CPU (%)
--- a/pkg/config/app_config.go
+++ b/pkg/config/app_config.go
@@ -53,8 +53,7 @@ type UserConfig struct {
 	// OS determines what defaults are set for opening files and links
 	OS OSConfig `yaml:"oS,omitempty"`
 
-	// Update is currently not being used, but like lazydocker, it may be used down
-	// the line to help you update automatically.
+	// UpdateConfig determines what the default settings are for updating the ui
 	Update UpdateConfig `yaml:"update,omitempty"`
 
 	// Stats determines how long lazydocker will gather container stats for, and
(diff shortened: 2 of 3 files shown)


dawidd6:
Maybe instead of `Time` suffix, use `Interval`. Seems to be more accurate for me.

docs/Config.md
```diff
@@ -53,6 +53,9 @@ oS:
   openCommand: open {{filename}}
   openLinkCommand: open {{link}}
 update:
+  refreshProjectTime: 100ms
+  refreshContainersAndServicesTime: 100ms
+  refreshVolumesTime: 100ms
   method: never
```
jesseduffield: may as well remove this line given that we're not going to use it

docs/Config.md
```diff
@@ -53,6 +53,9 @@ oS:
   openCommand: open {{filename}}
   openLinkCommand: open {{link}}
 update:
+  refreshProjectTime: 100ms
```
jesseduffield: Given that a user is unlikely to know which of these if any are causing performance issues, I wonder if we should just have a single duration. This also will make the code easier to maintain, because if we add a new thing that refreshes frequently we can just use the config value.

pkg/gui/gui.go
```diff
@@ -267,13 +267,34 @@ func (gui *Gui) Run() error {
 		gui.waitForIntro.Add(1)
 	}
 
+	updates := gui.Config.UserConfig.Update
+	projectDuration := time.Millisecond * 100
```
jesseduffield: I would set the defaults in app_config rather than here. We're currently doing it for max duration of stats collection, but we could clean that up by doing:
```
func mustParseDuration(durationStr string) time.Duration {
	duration, err := time.ParseDuration(durationStr)
	if err != nil {
		panic(err)
	}
	return duration
}
```

and then doing something like:
```
		Update: UpdateConfig{
			RefreshRate: mustParseDuration("100ms"),
		},
```

docs/Config.md
mjarkk: Jup i was thinking that too, but i was not sure if that was a good idea because the issue this solves probably only needs to change the volumes pull rate.  

But because of the *probably* in the sentence above and your point about maintainability i'll change it to a generic rule.

docs/Config.md
mjarkk: Why remove this if md file shows all config items (to my understanding).  
this file is easy to understand for new users or for users that want to quickly know if something explicit can be configured, this way they they don't have to search inside the godoc for that item.

pkg/gui/gui.go
mjarkk: This is indeed cleaner, i'll change it to that.

docs/Config.md
jesseduffield: sorry what I meant was: we're not using this 'method' key anywhere in the app. This is a vestige from lazygit where you could enable regular auto-updates (as in installing the latest lazygit), but we're not using it here so there's no harm in deleting it given that the context of the 'update' config key has changed from installations to refresh rates.

docs/Config.md
jesseduffield: sounds good :) If it turns out we want to be more granular, we can do that down the line

docs/Config.md
mjarkk: Fixed it

mjarkk:
@jesseduffield & @dawidd6 i've fixed the suggestions.

pkg/config/app_config.go
```diff
@@ -191,20 +191,12 @@ type OSConfig struct {
...
-	RefreshContainersAndServicesTime string `yaml:"refreshContainersAndServicesTime,omitempty"`
-
-	// RefreshVolumesTime determines the time betweens updates of the volumes panel
-	RefreshVolumesTime string `yaml:"refreshVolumesTime,omitempty"`
-	Method             string `yaml:"method,omitempty"`
+	// RefreshProjectTime determines the time betweens updates of all continues docker commands like docker ps, docker images, etc.
+	// It expects a valid duration like: 100ms, 2s, 200ns
+	// for docs see: https://golang.org/pkg/time/#ParseDuration
+	DockerRefreshInterval time.Duration `yaml:"dockerRefreshInterval,omitempty"`
+	Method                string        `yaml:"method,omitempty"`
```
jesseduffield: could we also remove this line given we're not using Method anywhere

pkg/config/app_config.go
```diff
@@ -356,10 +348,8 @@ func GetDefaultConfig() UserConfig {
 		},
 		OS: GetPlatformDefaultConfig(),
 		Update: UpdateConfig{
-			RefreshProjectTime:               "100ms",
-			RefreshContainersAndServicesTime: "100ms",
-			RefreshVolumesTime:               "100ms",
-			Method:                           "never",
+			DockerRefreshInterval: time.Millisecond * 100,
+			Method:                "never",
```
jesseduffield: here too

mjarkk:
@jesseduffield fixed the suggestions.

robinpecha:
I can't find documentation for this setting anywhere. How to reduce the interval? Lazydocker 0.23.3 temporarily uses the CPU at 100% after starting, after a few minutes it calms down.

yboulkaid:
Unfortunately it seems like [we're back to hardcoded values](https://github.com/jesseduffield/lazydocker/blob/bedde4a03791e678c510b98d0de8b0c2d23d8efa/pkg/gui/gui.go#L268-L270) :(

Issue #31, opened 2019-07-01
Scrolling beyond the most recent log should not be possible


ed-bassett:
Steps for reproduction:
1. Select the `Logs` tab in the right panel
1. Scroll in the direction of the more recent log messages using either the keyboard shortcuts or the mouse.
1. Keep scrolling after the most recent message reaches the top of the screen.

Desired behaviour:
It should not be possible to scroll past the most recent log message. Probably the best option would be to limit scrolling so that the most recent message cannot move further up than the bottom of the screen, but it would also be reasonable enough to limit it so that it can't move further up than the top of the screen.

mjarkk:
This is a problem on ever tab and it also effects scrolling side ways using the arrow keys.

jesseduffield:
This is actually not too tricky to fix, all the code is in `scrollUpMain` and `scrollDownMain` in `pkg/gui/main_panel.go`. I'll chuck a 'good first issue' label on this

jesseduffield:
This is done!

jesseduffield:
Also hi @ed-bassett

PR #130, opened 2019-07-16
scrolling does not go beyond the displayed text #31

winhung:
Feel free to comment on what can be improved.
I may have missed out something.
Thank you.


pkg/gui/main_panel.go
```diff
@@ -39,7 +51,25 @@ func (gui *Gui) scrollRightMain(g *gocui.Gui, v *gocui.View) error {
 	mainView := gui.getMainView()
 	ox, oy := mainView.Origin()
 
-	return mainView.SetOrigin(ox+gui.Config.UserConfig.Gui.ScrollHeight, oy)
+	content := mainView.BufferLines()
+	var largestNumberOfCharacters int
+	for _, txt := range content {
```
winhung: I don't like this loop as it is unnecessary to go thru it every single time.
If there's some function that's already built-in that can do something similar only once, please point it out to me.
Thank you.

winhung:
@jesseduffield Any chance of reviewing this ?

pkg/gui/main_panel.go
jesseduffield: Yeah I share your concern, this could be quite an expensive loop to run. What if we used `ViewBufferLines` instead so that we're only looping through the lines that are displayed? That way there will be far less to loop through

winhung: Hmm...i am not sure how that would help. ViewBufferLines returns an []string too and i would most probably have to loop thru each to find the longest line. Similar to what i have already implemented.
I am going to need your help pointing out what i have misunderstood :)

jesseduffield: The difference is in the length of the array. Sometimes we are dealing with diffs that are thousands of lines long, in which case iterating through the entire array of BufferLines will be very expensive. On the other hand, iterating through just the ViewBufferLines is quite cheap because it's limited to the height of the window.

winhung: Ok i understand. Changes coming soon

winhung:
@jesseduffield My PR is ready for review :)

pkg/gui/main_panel.go
```diff
@@ -39,7 +51,25 @@ func (gui *Gui) scrollRightMain(g *gocui.Gui, v *gocui.View) error {
...
+		if len(txt) > largestNumberOfCharacters {
+			largestNumberOfCharacters = len(txt)
+		}
+	}
+
+	scrollCount := gui.Config.UserConfig.Gui.ScrollHeight
+	sizeX, _ := mainView.Size()
+	if ox+sizeX >= largestNumberOfCharacters {
+		if largestNumberOfCharacters > sizeX {
```
jesseduffield: I'm not sure why we need this extra condition here. I would have thought that if ` ox+sizeX >= largestNumberOfCharacters` then we could just `return nil`

pkg/gui/main_panel.go
```diff
@@ -22,9 +22,21 @@ func (gui *Gui) scrollDownMain(g *gocui.Gui, v *gocui.View) error {
 		_, sy := mainView.Size()
 		y += sy
 	}
+
+	_, sizeY := mainView.Size()
+	scrollCount := gui.Config.UserConfig.Gui.ScrollHeight
+	totalLines := mainView.LinesHeight()
```
jesseduffield: Using LinesHeight causes an issue when we have wrapping turned on with:
```
gui:
  wrapMainPanel: true
```
The issue is that you can press down and actually end up scrolling up because of the mismatch between the total original of lines and the number of lines when wrapping is done.

I've just pushed a commit to the gocui master branch which adds a `ViewLinesHeight()` function that you can use instead and that should fix that bug :)

If you need help bumping the gocui version let me know :)

winhung: Hi there, do share with me what should i do to update the version. Thank you.

pkg/gui/main_panel.go
winhung: Yes, thank you for pointing that out. This is so much better.

pkg/gui/main_panel.go
winhung: Done ! I got it updated :)

winhung:
@jesseduffield Ready for review again :)

pkg/gui/main_panel.go
```diff
@@ -22,9 +22,17 @@ func (gui *Gui) scrollDownMain(g *gocui.Gui, v *gocui.View) error {
...
+	scrollCount := gui.Config.UserConfig.Gui.ScrollHeight
+	totalLines := mainView.ViewLinesHeight()
+	if oy+sizeY >= totalLines {
+		return nil
+	}
+
 	// for some reason we can't work out whether we've hit the bottomq
 	// there is a large discrepancy in the origin's y value and the length of BufferLines
-	return mainView.SetOrigin(ox, oy+gui.Config.UserConfig.Gui.ScrollHeight)
+	return mainView.SetOrigin(ox, oy+scrollCount)
```
winhung: scrollCount is no longer needed. I will put the original change back.
Same for line 58.

winhung: Done

pkg/gui/main_panel.go
```diff
@@ -22,6 +22,13 @@ func (gui *Gui) scrollDownMain(g *gocui.Gui, v *gocui.View) error {
 		_, sy := mainView.Size()
 		y += sy
 	}
+
+	_, sizeY := mainView.Size()
+	totalLines := mainView.ViewLinesHeight()
+	if oy+sizeY >= totalLines {
+		return nil
+	}
```
jesseduffield: I found a couple extra issues with this while testing. Mostly around the `gui.Config.UserConfig.Gui.ScrollPastBottom` flag. When that's set we should be able to continue scrolling until the main panel's buffer is empty. This minor alteration fixed the issues that I found :)

```
func (gui *Gui) scrollDownMain(g *gocui.Gui, v *gocui.View) error {
	mainView := gui.getMainView()
	mainView.Autoscroll = false
	ox, oy := mainView.Origin()

	reservedLines := 0
	if !gui.Config.UserConfig.Gui.ScrollPastBottom {
		_, sizeY := mainView.Size()
		reservedLines = sizeY
	}
	totalLines := mainView.ViewLinesHeight()
	if oy+reservedLines >= totalLines {
		return nil
	}

	return mainView.SetOrigin(ox, oy+gui.Config.UserConfig.Gui.ScrollHeight)
}
```

winhung: Oh, i didn't realize there was such an option hahaha.
Ok, will put that in.

winhung:
> 
> 
> @jesseduffield Done and ready for review :)

@jesseduffield Gentle reminder

jesseduffield:
Fantastic work! Thanks for making this

PR #136, opened 2019-07-23
Correlate Services and Containers by Service Name

### LINKED ISSUE #122: Service and Container correlation is broken
**Describe the bug**
When multiple projects use the same service names, it's possible – and potentially common – for the listed services to refer to another project's containers.

**To Reproduce**
1. Create two separate `docker-compose.yml` files in two different directories.
2. Add both an `app` service and a project-specific service to each `docker-compose.yml` file.
3. Build and run both project's containers.
4. Run `lazydocker` in both projects.

**Expected behaviour**
Each instance of `lazydocker` shows two services, and both services refer to different containers.

**Observed behavior**
`lazydocker` matches the Service record against the first Container with a matching service name label.  This can be easily validated by examining the `com.docker.compose.project` label in the "Container Config" panel for the listed services.

This bug has a few knock-on effects:

* The Containers panel only shows containers that have a service name that does not match a Service in the current project.  If (e.g.) the `app` Service has captured the container for the wrong project, the container for the *correct* `app` Service is also inaccessible.
* Since the `docker-compose` project name cannot be reliably determined externally (without introducing a new convention), `lazydocker` currently relies on the appropriate label of the first Service's container to determine the project name (falling back to `$PWD`).  If that Service has captured the container for the wrong project, the displayed project name will also be incorrect.

 - OS: macOS 10.14.5
 - Lazydocker v0.5.5

**Additional context**

The Service and Container matching behavior described is supported by a reading of the current logic here: https://github.com/jesseduffield/lazydocker/blob/bd16a78b06542bb956d869e4d0285188fe6353e9/pkg/commands/docker.go#L249-L290

pvande:
The config hash ends up not being a unique identifier of a Service.  Additionally, Services may have multiple associated Containers.

``` yaml
# Sample docker-compose.yml
version: "2.2"

services:
  named-service:
    container_name: 'named-container'
    image: 'alpine'
    command: 'sh -c "while :; do sleep 1; done"'

  anonymous-service:
    image: 'alpine'
    command: 'sh -c "while :; do sleep 1; done"'

  scaled-service:
    image: 'alpine'
    scale: 3
    command: 'sh -c "while :; do sleep 1; done"'
```

``` bash
$ docker-compose config --hash="*"
named-service e6e7c4564dfc435ba96725098c62e6b6cf8da9f0b76b0a08be2baddedad6d007
anonymous-service 254f4815ef36bc9fcc2ce80f82278187c9e233309a4106951551f3692657ce9c
scaled-service 254f4815ef36bc9fcc2ce80f82278187c9e233309a4106951551f3692657ce9c
```

This commit changes how Services are discovered (`docker-compose config --services`), how they're associated (by Container ID, rather than config hash or name), and allows for Services to reference multiple Containers (though only the first is directly used for anything at this time).

Resolves #122.

### DIFF
--- a/pkg/commands/docker.go
+++ b/pkg/commands/docker.go
@@ -216,20 +216,26 @@ func (c *DockerCommand) RefreshContainersAndServices() error {
 		}
 	}
 
-	c.assignContainersToServices(containers, services)
-
 	var displayContainers = containers
-	if !c.Config.UserConfig.Gui.ShowAllContainers {
-		displayContainers = c.obtainStandaloneContainers(containers, services)
+
+	if c.InDockerComposeProject {
+		var standaloneContainers, err = c.assignContainersToServices(containers, services)
+		if err != nil {
+			return err
+		}
+
+		if !c.Config.UserConfig.Gui.ShowAllContainers {
+			displayContainers = standaloneContainers
+		}
 	}
 
 	// sort services first by whether they have a linked container, and second by alphabetical order
 	sort.Slice(services, func(i, j int) bool {
-		if services[i].Container != nil && services[j].Container == nil {
+		if len(services[i].Containers) > 0 && len(services[j].Containers) == 0 {
 			return true
 		}
 
-		if services[i].Container == nil && services[j].Container != nil {
+		if len(services[i].Containers) == 0 && len(services[j].Containers) > 0 {
 			return false
 		}
 
@@ -243,17 +249,40 @@ func (c *DockerCommand) RefreshContainersAndServices() error {
 	return nil
 }
 
-func (c *DockerCommand) assignContainersToServices(containers []*Container, services []*Service) {
-L:
-	for _, service := range services {
-		for _, container := range containers {
-			if !container.OneOff && container.ServiceName == service.Name {
-				service.Container = container
-				continue L
-			}
-		}
-		service.Container = nil
+func (c *DockerCommand) assignContainersToServices(containers []*Container, services []*Service) ([]*Container, error) {
+	// Get a list of the Container IDs for this Docker Compose project.
+	composeCommand := c.Config.UserConfig.CommandTemplates.DockerCompose
+	output, err := c.OSCommand.RunCommandWithOutput(fmt.Sprintf("%s ps -q", composeCommand))
+	if err != nil {
+		return nil, err
 	}
+
+	// Create an index for Container IDs to make lookups less expensive.
+	lines := utils.SplitLines(output)
+	serviceContainerIds := make(map[string]bool)
+	for _, line := range lines {
+		serviceContainerIds[line] = true
+	}
+
+	// Index Services by name.
+	serviceMap := make(map[string]*Service)
+	for _, service := range services {
+		serviceMap[service.Name] = service
+	}
+
+	// Walk the list of Containers, categorizing them as either associated with a
+	// Service or as standalone.
+	standaloneContainers := make([]*Container, 0, len(containers) - len(serviceContainerIds))
+	for _, container := range containers {
+		if !container.OneOff && serviceContainerIds[container.ID] {
+			service := serviceMap[container.ServiceName]
+			service.Containers = append(service.Containers, container)
+		} else {
+			standaloneContainers = append(standaloneContainers, container)
+		}
+	}
+
+	return standaloneContainers, nil
 }
 
 // filterOutExited filters out the exited containers if c.ShowExited is false
@@ -268,22 +297,6 @@ func (c *DockerCommand) filterOutExited(containers []*Container) []*Container {
 		}
 	}
 	return toReturn
-}
-
-// obtainStandaloneContainers returns standalone containers. Standalone containers are containers which are either one-off containers, or whose service is not part of this docker-compose context
-func (c *DockerCommand) obtainStandaloneContainers(containers []*Container, services []*Service) []*Container {
-	standaloneContainers := []*Container{}
-L:
-	for _, container := range containers {
-		for _, service := range services {
-			if !container.OneOff && container.ServiceName != "" && container.ServiceName == service.Name {
-				continue L
-			}
-		}
-		standaloneContainers = append(standaloneContainers, container)
-	}
-
-	return standaloneContainers
 }
 
 // GetContainers gets the docker containers
@@ -349,25 +362,20 @@ func (c *DockerCommand) GetServices() ([]*Service, error) {
 	}
 
 	composeCommand := c.Config.UserConfig.CommandTemplates.DockerCompose
-	output, err := c.OSCommand.RunCommandWithOutput(fmt.Sprintf("%s config --hash=*", composeCommand))
+	output, err := c.OSCommand.RunCommandWithOutput(fmt.Sprintf("%s config --services", composeCommand))
 	if err != nil {
 		return nil, err
 	}
 
-	// output looks like:
-	// service1 998d6d286b0499e0ff23d66302e720991a2asdkf9c30d0542034f610daf8a971
-	// service2 asdld98asdklasd9bccd02438de0994f8e19cbe691feb3755336ec5ca2c55971
-
-	lines := utils.SplitLines(output)
-	services := make([]*Service, len(lines))
-	for i, str := range lines {
-		arr := strings.Split(str, " ")
+	serviceNames := utils.SplitLines(output)
+	services := make([]*Service, len(serviceNames))
+	for i, name := range serviceNames {
 		services[i] = &Service{
-			Name:          arr[0],
-			ID:            arr[1],
+			Name:          name,
 			OSCommand:     c.OSCommand,
 			Log:           c.Log,
 			DockerCommand: c,
+			Containers:    make([]*Container, 0, 1),
 		}
 	}
 
--- a/pkg/commands/docker.go
+++ b/pkg/commands/docker.go
@@ -272,7 +272,9 @@ func (c *DockerCommand) assignContainersToServices(containers []*Container, serv
 
 	// Walk the list of Containers, categorizing them as either associated with a
 	// Service or as standalone.
-	standaloneContainers := make([]*Container, 0, len(containers) - len(serviceContainerIds))
+	serviceContainerCount := len(serviceContainerIds)
+	standaloneContainerCount := len(containers) - serviceContainerCount
+	standaloneContainers := make([]*Container, 0, standaloneContainerCount)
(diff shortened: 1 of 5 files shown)


pkg/commands/docker.go
```diff
@@ -349,25 +365,20 @@ func (c *DockerCommand) GetServices() ([]*Service, error) {
...
+	services := make([]*Service, len(serviceNames))
+	for i, name := range serviceNames {
 		services[i] = &Service{
-			Name:          arr[0],
-			ID:            arr[1],
+			Name:          name,
 			OSCommand:     c.OSCommand,
 			Log:           c.Log,
 			DockerCommand: c,
+			Containers:    make([]*Container, 0, 1),
```
jesseduffield: why are we setting a capacity here?

pkg/commands/service.go
```diff
@@ -13,27 +13,26 @@ import (
...
 // GetDisplayStrings returns the dispaly string of Container
 func (s *Service) GetDisplayStrings(isFocused bool) []string {
 
-	if s.Container == nil {
+	if len(s.Containers) == 0 {
 		return []string{utils.ColoredString("none", color.FgBlue), s.Name, ""}
 	}
 
-	cont := s.Container
+	cont := s.Containers[0]
```
jesseduffield: given how often we're using this `s.Container[0]` approach, could we factor that out into a function like
```
func (s *Service) SelectedContainer() Container {
	return s.Containers[0]
}
```

That way, when we add the code for dealing with multiple containers, we only need to change it in one place

pkg/commands/docker.go
```diff
@@ -243,17 +249,43 @@ func (c *DockerCommand) RefreshContainersAndServices() error {
...
+
+	// Index Services by name.
+	serviceMap := make(map[string]*Service)
 	for _, service := range services {
-		for _, container := range containers {
-			if !container.OneOff && container.ServiceName == service.Name {
-				service.Container = container
-				continue L
-			}
+		service.Containers = make([]*Container, 0, 1)
```
jesseduffield: why are we setting a capacity of 1 here?

pkg/commands/docker.go
```diff
@@ -243,17 +249,43 @@ func (c *DockerCommand) RefreshContainersAndServices() error {
...
+	}
+
+	// Walk the list of Containers, categorizing them as either associated with a
+	// Service or as standalone.
+	serviceContainerCount := len(serviceContainerIds)
+	standaloneContainerCount := len(containers) - serviceContainerCount
+	standaloneContainers := make([]*Container, 0, standaloneContainerCount)
+	for _, container := range containers {
+		if !container.OneOff && serviceContainerIds[container.ID] {
+			service := serviceMap[container.ServiceName]
```
jesseduffield: is it possible here that a container may not have a ServiceName set against it? If so, we should add a check for that so that we don't get a nil pointer exception when we say `service.Containers`

pkg/commands/service.go
pvande: We certainly can, though I'm not sure how useful that would be overall.  `commands.Service` doesn't really have `gui` responsibilities, so delegating "selection" responsibilities to it seems to conflate concerns a bit.  A function that *did* have `gui` responsibilities might look like

``` golang
func (gui *Gui) selectedContainerForService(s *Service) *Container {
	idx := gui.State.Panels.Services.SelectedContainer[s.ID]
	return s.Containers[idx]
}
```

I'm not sure how relevant I would consider this for dealing with multiple containers, though.  The majority of the `s.Containers[0]` cases are treating the Service as a proxy for the Container, not acting on a "selection", and when we have the ability to select Containers for a Service that behavior doesn't make sense – we should be able to treat that selection exactly like any other Container.  At that point, Services need to stop being proxies for Containers, and start behaving like groups of Containers (e.g. "Logs" show the interleaved logs from all containers for the Service, "Container Config" goes away, "Stats" either presents an aggregate view or goes away, etc.).

A reasonable case could be made that when only a single Container exists for a Service, the two should be treated interchangeably.  In my own explorations around that, I've found having a clear delineation between Service and Container in the UI helps both predictability and implementation, and provides a transparent solution to other issues for free (e.g. the implicit vs explicit container name issue goes away if you're not collapsing Services and Containers into the same item).

jesseduffield: yeah I agree. If I recall correctly, I did use `docker-compose logs <service name>` being used at some point but decided it was a bit jarring to see the service name at the side when it was obvious from the context what service you were looking at, so I just showed the container logs instead.

I agree about having a clear separation between the two, and I'm happy for that to be bigger discussion down the road. At any rate, feel free to use that `selectedContainerForService` function if you wish, I'll defer to your judgement.

pkg/commands/docker.go
jesseduffield: fair enough

jesseduffield:
Code looks good :) After a functional test, I've realised that when you remove a service's container, you momentarily see it appear as an exited container in the containers panel, even if it was already exited before you removed it. Is this expected behaviour?

pkg/commands/service.go
pvande: Absolutely – in the current UI, with the current implicit tie between a Service and a single Container, it would absolutely feel bad to see the container name next to every log line.  Container logs make perfect sense in that context; the issue comes from the fact that we're going to need to leave that context behind.

For now, I'm content to leave this as `service.Containers[0]` since it's both temporary (🤞) and the simplest translation of the existing logic.

pvande:
> Code looks good :) After a functional test, I've realised that when you remove a service's container, you momentarily see it appear as an exited container in the containers panel, even if it was already exited before you removed it. Is this expected behaviour?

I hadn't seen that, but I can't say I'm surprised.  This change would remove the service association as soon as `docker-compose ps` stops returning the container ID...  I can't quite tell whether it loses the label because Docker Compose removes it or because Docker stops reporting labels for things it's actively removing, but I _believe_ it's the latter.

Whether that's an idiosyncrasy that should be exposed to users of `lazydocker` is a separate decision.  (The work-around, I suppose, would involve manually retaining any Service-aligned Containers that were in a state of "removing".)

jesseduffield:
Hmm it's a tricky condundrum because though I try to keep things fairly stateless, it is a bad user experience to see the container jump in and out of the container panel whenever it gets removed. I think we should find a way to conceal it before we merge this PR.

I've just done some more functional testing and I've found that if you have no containers running at all, and then you hit 'r' on a service to restart it, the application crashes because `len(containers)` is still zero for a bit while `serviceContainerCount` is 1, meaning we try to initialize `standaloneContainers` with a capacity of -1. Seems like this may be a similar issue to the above issue in that there is a mismatch in what docker and docker-compose returns about containers. The docker-compose `ps` command takes 600ms, the GetContainer() function takes 3ms, so there's a window for things to get out of sync. I tried swapping around the order in which those functions are called, to no avail. I think we have a bit of a race condition issue where if you're removing a service or restarting one, there's an opportunity for the two requests to come back with mismatched data.

Currently, if I remove all standalone containers and go through my services and remove till there's nothing left, or start from no services and start restarting them, I intermittently (though fairly reliably) get crashes.

pvande:
Here are my observations:

* I cannot reproduce your error around restarting services without containers.
* On my machine, `docker-compose ps -q` takes ~450ms to run.  `RefreshContainersAndServices` is scheduled to run (by default) every 100ms.  At the very least, this leads to a long queue for the `ServiceMutex` that only ever grows.
* Scheduling `RefreshContainersAndServices` to run every 1s "solves" both problems, in that it both affords time for the function to return before being scheduled again and that it is far less likely to catch the fractional second where Docker is aware of the container but `docker-compose ps` doesn't list it.  (This is admittedly an unsatisfying fix.)
* As for the race condition between Container and Service data sources, I'd love to find an alternative solution that's faster (and can potentially run in the same process).  All of the data that we need to do correlation is in the Container labels, so it's theoretically possible.  The problem is, once again, Docker Compose projects.  The project label is populated by Docker Compose and used internally by Docker Compose.  Unfortunately, the project name is not externalized by Docker Compose, and can (currently) be configured by a command line argument, an environment variable, a `.env` file, and/or the directory name (all of which are then subjected to string normalization, and new options are being discussed and implemented).
  * This leaves our "best" option as follows:
    1.  Run `docker-compose ps -q` to grab Container IDs related to this project.
    2.  Fetch the Container list.
    3.  Fetch the list of Service names.
    3.  Grab the project name from the first Container ID that exists in the Container list.
    4.  Filter the Container set to those with a matching project name.
    5.  Group the matching Containers by Service name.
  * Brief experimentation suggests this will also solve the "jumping" service we've witnessed elsewhere.

jesseduffield:
I should clarify what I said earlier about the crashing: I get reliable crashes when I have removed all the containers (i.e. docker-compose down, and removing any other containers outside the docker compose project) and I restart a service. The crash happens because the length of `containers` is less than the length of `serviceContainerIds` because of the following:
1) we obtain `containers` (~3ms) 
2) the user adds the new container by starting the service
3) we obtain `serviceContainerIds` (~500ms) 
4) we set `standaloneContainerCount` based on the difference in length of the two arrays, and it ends up negative.
5) We try to instantiate `standaloneContainers` with a negative capacity.

If we don't really want to rely on container labels as you say, we can actually work around this by:
1) obtaining `containers` after obtaining `serviceContainerIds` (meaning now the race condition issue is only 3ms rather than 500ms)
2) creating `standaloneContainers` like so:
```
standaloneContainers := []*Containers{}
```
This is admittedly less efficient than setting the capacity ahead of time but it's a small price to pay given we're only going to have 10 or so containers.

Functionally testing this approach, we don't have any crashes and we don't see things jumping into the containers panel while being removed.

I've branched off this branch and added a commit showing functionally how to do this: `PR-136-crash-fixes` (commit: `c435185`).

I think this solves the problem. There is still a race condition, but it's inconsequential now that we're not making hard assertions on what both requests are returning.

As for the issue of the 100ms loop period causing a queue on the mutex, this is actually not a problem because `time.Tick` drops ticks when the receiver's duration exceeds the tick duration.

I think I've addressed your comments. Let me know your thoughts :)

wesselvdv:
@pvande @jesseduffield Is there any animo to pursue this PR into lazydocker? I have been hit by this issue for quite some time now.

pvande:
> @pvande @jesseduffield Is there any animo to pursue this PR into lazydocker?

I’m sorry; my available time for OSS development work has dropped off significantly, so this is not something I’m able to pursue at this time.

jesseduffield:
Closing due to staleness

PR #138, opened 2019-07-25
Fix image reference

tobypinder:
According to rest of docs this should be "lazyteam"


dawidd6:
Yep, you're right. Thanks!

Issue #144, opened 2019-08-02
add image name to container frame


anjiawei1991:
![2019-08-02 10-15-14屏幕截图](https://user-images.githubusercontent.com/48538419/62339904-3ed8a380-b510-11e9-99b3-271ea6ce21cf.png)

I think the image name should be added to the container frame because it's a really important thing especially when I haven't named the container such as the `nice_cartwright` and `angry_chatterjee` in the above picture. And, I can't see the image name in any of the right pages.

anjiawei1991:
![2019-08-02 10-57-23屏幕截图](https://user-images.githubusercontent.com/48538419/62341058-62055200-b514-11e9-9958-e5c5c5eaa5ec.png)

I have done that job, and it looks nice I think.

mjarkk:
Have you also tried looking at it in a small terminal window?  
I think the 0.00% is going to out of the window, and the cpu usage is for me more important than the image names.  

The problem with these wide lines is that not everyone has there terminal maximized and for those these these changes look wrong.  
For example when i start lazydocker these changes will look like this:  
```
|running nice_cartwright  fortu|
|created angry_chatterjee f14d3|
|running nsq_nsqd_1       nsqio| <- this container name is now completly useless because i only see the username of who made the container

-> compaired to this
|running nice_cartwright  0.14%|
|created angry_chatterjee 0.00%|
|running nsq_nsqd_1       5.50%|
```

anjiawei1991:
![2019-08-02 15-05-55屏幕截图](https://user-images.githubusercontent.com/48538419/62351056-1617d480-b537-11e9-9430-5bef4afc2a3a.png)


@mjarkk It works well if window's width is greater than 1150. CPU is indeed important. So another way is to place the image name behind the CPU. Or more sophisticated, decide wether to show the image name depend on the  window's size, but it will bring complexity. 

But I do think the image name is very important. I always use the terminal in max window size mode, and I think most of users do this.

jesseduffield:
I've got no issue with showing the image name after the cpu.

Although, it seems that there is a strong use case for being able to see a lot of things related to a panel list item in the side panel. Perhaps we should consider having a toggle-able expanded side panel view for when you want to see extra information about all your e.g. containers at once.

At any rate for this issue I'm happy for the image name to go after the cpu.

anjiawei1991:
@jesseduffield 
It sounds good to have a toggle-able expanded side panel view and that could be another issue.

[...]

PR #145, opened 2019-08-03
add image to conatiner view.

anjiawei1991:
Fix: https://github.com/jesseduffield/lazydocker/issues/144

### DIFF
--- a/pkg/commands/container.go
+++ b/pkg/commands/container.go
@@ -248,7 +248,9 @@ type ContainerCliStat struct {
 
 // GetDisplayStrings returns the dispaly string of Container
 func (c *Container) GetDisplayStrings(isFocused bool) []string {
-	return []string{c.GetDisplayStatus(), c.Name, c.GetDisplayCPUPerc()}
+	image := strings.TrimPrefix(c.Container.Image, "sha256:")
+
+	return []string{c.GetDisplayStatus(), c.Name, c.GetDisplayCPUPerc(), utils.ColoredString(image, color.FgMagenta)}
 }
 
 // GetDisplayStatus returns the colored status of the container


pkg/commands/container.go
```diff
@@ -248,7 +248,11 @@ type ContainerCliStat struct {
 
 // GetDisplayStrings returns the dispaly string of Container
 func (c *Container) GetDisplayStrings(isFocused bool) []string {
-	return []string{c.GetDisplayStatus(), c.Name, c.GetDisplayCPUPerc()}
+	image := c.Container.Image
```
jesseduffield: we can just do `image := strings.TrimPrefix(c.Container.Image, "sha256:")` here

pkg/commands/container.go
```diff
@@ -248,7 +248,11 @@ type ContainerCliStat struct {
...
+	if strings.HasPrefix(image, "sha256:") {
+		image = image[7:20]
+	}
+	return []string{c.GetDisplayStatus(), c.Name, c.GetDisplayCPUPerc(), image}
```
jesseduffield:
```suggestion
	return []string{c.GetDisplayStatus(), c.Name, c.GetDisplayCPUPerc(), utils.ColoredString(image, color.FgMagenta)}
```
What are your thoughts on giving this a color other than the default? I think that makes the CPU percentage easier to read:
image

pkg/commands/container.go
anjiawei1991: 2019-08-06 19-01-05屏幕截图

Since the image name may remain too long, I have just reserved 14 characters here.

anjiawei1991: 2019-08-06 19-08-00屏幕截图

Compare with this one which is more concise and won't take much spaces.

jesseduffield: I personally don't mind if the image name is long, given we don't yet have any other columns to the right, and the magenta color means it's not too jarring on the eyes. But if you want to shorten it, I would add an ellipsis (`...`) after the 14 characters to communicate to the user that there's more of the image name not being shown

anjiawei1991: I think long name is not very good. But I do think that add an ellipsis after the 14 characters is better than before.  I'm going to do it.

anjiawei1991: If I do that,  it would be inconsistent with those named containers whose name is very long and with the other colums such as container name.

It may not be that bad to remain the whole name since there's no colums to the right now.

Though, I think it could be adjusted in the future when  adding  toggle-able expanded side panel view.

anjiawei1991: So I'll adopt your suggestions that remain the whole name after `sha256:`

anjiawei1991: @jesseduffield all is done. thank you.

jesseduffield:
Merged! Thanks for making this :)

Issue #147, opened 2019-08-05
lazydocker is spawning orphaned docker logs processes


keigel2001:
**Describe the bug**
Everytime I open lazydocker, it's spwaning new processes of `docker logs`, which never get closed when closing lazydocker. After every usage of lazydocker, there's more processes.

**To Reproduce**
I'm using the binary release v0.6.4 of lazydocker and have one docker container running.
Open lazydocker and close it.
Then check f.e. with `ps aux | grep logs` for `docker logs`-processes.

**Expected behaviour**
`docker logs` processes get closed, after closing lazydocker.

**Screenshots**
![image](https://user-images.githubusercontent.com/9857738/62442595-aaad4d00-b758-11e9-8b96-ea09536a9418.png)

**Desktop (please complete the following information):**
 - OS: `Linux debian 4.19.0-5-amd64 #1 SMP Debian 4.19.37-5+deb10u1 (2019-07-19) x86_64 GNU/Linux`
also happenes with debian stretch (`Linux debian 4.9.0-9-amd64 #1 SMP Debian 4.9.168-1+deb9u4 (2019-07-19) x86_64 GNU/Linux`)
 - Docker: `19.03.1, build 74b1e89`
 - Lazydocker Version v0.6.4 (Binary Release)

glvr182:
In container_panel.go we handle view logs like this:
```go
func (gui *Gui) handleContainerViewLogs(g *gocui.Gui, v *gocui.View) error {
	container, err := gui.getSelectedContainer()
	if err != nil {
		return nil
	}

	c, err := container.ViewLogs()
	if err != nil {
		return gui.createErrorPanel(gui.g, err.Error())
	}

	gui.SubProcess = c
	return gui.Errors.ErrSubProcess
}
```
As we can see this calls container.ViewLogs(), which returns a cmd:
```go
// ViewLogs attaches to a subprocess viewing the container's logs
func (c *Container) ViewLogs() (*exec.Cmd, error) {
	templateString := c.OSCommand.Config.UserConfig.CommandTemplates.ViewContainerLogs
	command := utils.ApplyTemplate(
		templateString,
		c.DockerCommand.NewCommandObject(CommandObject{Container: c}),
	)

	cmd := c.OSCommand.ExecutableFromString(command)
	c.OSCommand.PrepareForChildren(cmd)

	return cmd, nil
}
```
Then the returned command is used in a subprocess and run like this:
```go
func (gui *Gui) runCommand() error {
[...]

PR #148, opened 2019-08-06
ensure the current task is stopped before exit

jesseduffield:
fixes https://github.com/jesseduffield/lazydocker/issues/147 (kinda)

For some reason we have child processes that are not being killed when lazydocker exits. To fix this, I'm making it so that before closing, lazydocker kills its task manager, which in turn kills the current task if there is any (e.g. `docker logs <container id>`). If the child process isn't killed in three seconds, it prints an error to stdout.

I have no idea why the child processes are not being killed when the parent is stopped in the first place. I considered maybe their group id is different, given we assign a group id to some subprocesses like `docker-compose logs`, but it seems that's not the case with `docker logs`. I am open to suggestions :)

### DIFF
--- a/pkg/tasks/tasks.go
+++ b/pkg/tasks/tasks.go
@@ -1,9 +1,11 @@
 package tasks
 
 import (
+	"fmt"
 	"sync"
 	"time"
 
+	"github.com/jesseduffield/lazydocker/pkg/i18n"
 	"github.com/sirupsen/logrus"
 )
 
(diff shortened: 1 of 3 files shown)


pkg/tasks/tasks.go
```diff
@@ -24,8 +27,31 @@ type Task struct {
...
-	return &TaskManager{Log: log}
+func NewTaskManager(log *logrus.Entry, translationSet *i18n.TranslationSet) *TaskManager {
+	return &TaskManager{Log: log, Tr: translationSet}
+}
+
+// Close closes the task manager, killing whatever task may currently be running
+func (t *TaskManager) Close() {
+	return
+
+	if t.currentTask == nil {
```
golangcibot: unreachable: unreachable code (from `govet`)

PR #149, opened 2019-08-08
remove tap command for homebrew

ekremkaraca:
No longer needed, it's already in core formula

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ Memorising docker commands is hard. Memorising aliases is slightly less hard. Ke
 
 ### Homebrew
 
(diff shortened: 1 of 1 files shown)


dawidd6:
Yes, we are aware of this, but idk if it's a good idea to completely not mention the tap.
I mean it's already nicely configured and updates are almost instant, where in the core one must to manually submit a PR and wait for maintainers.

So I propose to mention that the tap exists and can be used to get updates quicker.

jesseduffield:
I agree with @dawidd6 :)

ekremkaraca:
@dawidd6 @jesseduffield I've updated my commit after your valuable comments.

dawidd6:
Nice, thanks man.

ekremkaraca:
@dawidd6 you're welcome:)

PR #150, opened 2019-08-12
fix german notion of volumes

Duske:
Afaik, `Volumen` is not the common notion for docker volumes. Simply using `Volumes` is better.


jesseduffield:
Thanks for making this :)

PR #156, opened 2019-08-25
Improve container top

skanehira:
Check container state.
If container is not running then not call top api.

image


dawidd6:
Nice PR, thanks!

Issue #158, opened 2019-08-29
Return from main panel after help menu


jakobkogler:
**Describe the bug**
After entering the main panel, and opening the help menu, it is no longer possible to exit the main panel with key bindings.

**To Reproduce**

1. Start `lazydocker`
2. Enter the main panel of the first container with `<ENTER>`
3. Open help menu with `x` or `?`
4. Close the help menu with `<ESC>`
5. Try to leave the main panel with `<ESC>`

The last command doesn't work, and the main panel stays focused. It also shows some other funky behaviors, like changing tabs doesn't work very well.

**Desktop (please complete the following information):**
 - OS: Ubuntu 18.10
 - Lazydocker VersionVersion:

```
    Version: 0.7.1
    Date: 2019-08-25T17:27:17Z
    BuildSource: binaryRelease
    Commit: febfce995c852ca9d1977febc6098165f7e31c8f
    OS: linux
    Arch: 386
```

dawidd6:
@jakobkogler I think this could be fixed in this [PR](https://github.com/jesseduffield/lazydocker/pull/159).

Can you build it from source and do some testing?

jakobkogler:
Hmm, somehow I can't get the app to build/run correctly.
I did:

```
docker build -t lazyteam/lazydocker \
    --build-arg BUILD_DATE=`date -u +"%Y-%m-%dT%H:%M:%SZ"` \
    --build-arg DOCKER_VERSION="v$(docker -v | cut -d" " -f3 | rev | cut -c 2- | rev)"\
    --build-arg VCS_REF=`git rev-parse --short HEAD` \
    --build-arg VERSION=`git describe --abbrev=0 --tag` \
    .
```

Both with and without the `DOCKER_VERSION` argument, and then tried to start it with 

```
docker run -it -v /var/run/docker.sock:/var/run/docker.sock <image>
```

I get the error message:

```
panic: interface conversion: interface {} is nil, not string

goroutine 1 [running]:
github.com/jesseduffield/lazydocker/pkg/gui.(*Gui).layout(0xc00036a420, 0xc00030e420, 0xc000304018, 0xc0005e9768)
        /tmp/gobuild/pkg/gui/layout.go:97 +0x1d05
github.com/jesseduffield/gocui.ManagerFunc.Layout(0xc0002cef70, 0xc00030e420, 0x0, 0x1)
        /tmp/gobuild/vendor/github.com/jesseduffield/gocui/gui.go:390 +0x30
[...]

PR #159, opened 2019-08-31
Implement PreviousViews stack

dawidd6:
Closes: #158 

~Needs more investigation, cause PreviousViews stack keeps growing  with time.~ should be ok now

### DIFF
--- a/pkg/gui/gui.go
+++ b/pkg/gui/gui.go
@@ -149,16 +149,6 @@ func NewGui(log *logrus.Entry, dockerCommand *commands.DockerCommand, oSCommand 
 		PreviousViews: stack.New(),
 	}
 
-	/*
-		// debugging PreviousViews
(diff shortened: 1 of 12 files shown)


jesseduffield:
Nice! Code looks good, and I gave it a functional test

pkg/gui/gui.go
```diff
@@ -144,9 +145,20 @@ func NewGui(log *logrus.Entry, dockerCommand *commands.DockerCommand, oSCommand
 			},
 			Project: &projectState{ContextIndex: 0},
 		},
-		SessionIndex: 0,
+		SessionIndex:  0,
+		PreviousViews: stack.New(),
 	}
 
+	/*
```
jesseduffield: if you wanna just remove this comment here, I'm happy for us to merge this :)

PR #162, opened 2019-09-03
Update keybindings.go

Tuurlijk:
Remove duplicate keybinding: volumes - c
Bind F10 to gui.quit

### DIFF
--- a/pkg/gui/keybindings.go
+++ b/pkg/gui/keybindings.go
@@ -67,12 +67,6 @@ func (gui *Gui) GetInitialKeybindings() []*Binding {
 		},
 		{
 			ViewName: "",
-			Key:      gocui.KeyF10,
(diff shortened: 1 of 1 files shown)


dawidd6:
Nice, but why F10 tho?

Tuurlijk:
Muscle memory from the Norton Commander / Midnight Commander days

glvr182:
I would say that this can be done by the users themselfs when #117 is merged. No need to make this into a hardcoded config

Tuurlijk:
Ok, but remove the duplicate volume - c keybinding.

dawidd6:
Can you make another commit removing this new binding? @Tuurlijk

glvr182:
> Ok, but remove the duplicate volume - c keybinding.

Yea I removed that in the PR.

PR #179, opened 2019-10-21
Remove the containers once you exit

Jamesking56:
Exiting after using `docker run` leaves behind stopped containers. Adding `--rm` tells Docker to delete them after exiting.


dawidd6:
Good catch, thanks.

PR #181, opened 2019-10-23
Removed unconditional logging profile

skovtunenko:
The `lazydocker` application will respect the given profile when it comes to logging.


dawidd6:
Thanks!

PR #182, opened 2019-10-28
Return stdout only from docker-compose command

### LINKED ISSUE #177: Empty lines in "Services" panel show words "none", "Some", "The"
**Describe the bug**
Empty lines in "Services" panel show words "none", "Some", "The".

**To Reproduce**
Install lazydocker 0.7.4 on Ubuntu 18.04, change directory to one containing a `docker-compose.yml` file, run lazydocker, observe "Services" panel.

**Expected behaviour**
Empty lines should be empty (contain no text)

**Screenshots**
This screenshot shows "Services" panel when there are no services with containers yet:
Screenshot showing  words "none", "Some", "The" in "Services" panel

This screenshot shows "Services" panel when there are 11 services. Note the two remaining lines containing  "none", "Some", "The"
Another screenshot showing  words "none", "Some", "The" in "Services" panel

 - OS: Ubuntu 18.04.3 LTS, Linux 5.0.0-25-generic
 - Docker version 19.03.3, build a872fc2f86
 - docker-compose version 1.24.1, build 4667896b
 - Lazydocker Version: 0.7.4

**Additional context**
I can reproduce it on multiple machines with similar specs/config

harrywhite4:
This fixes #177 which was caused by warning log messages being output from the `docker-compose config` command. These messages were also visible when viewing the docker-compse config and service tops.

Since the messages were being output to stderr i thought the best fix would be to only capture stdout from these commands. I did also keep the behaviour of outputting and error with stderr on failure.

### DIFF
--- a/pkg/commands/os.go
+++ b/pkg/commands/os.go
@@ -60,7 +60,7 @@ func (c *OSCommand) SetCommand(cmd func(string, ...string) *exec.Cmd) {
 func (c *OSCommand) RunCommandWithOutput(command string) (string, error) {
 	cmd := c.ExecutableFromString(command)
 	before := time.Now()
-	output, err := sanitisedCommandOutput(cmd.CombinedOutput())
+	output, err := sanitisedCommandOutput(cmd.Output())
 	c.Log.Warn(fmt.Sprintf("'%s': %s", command, time.Since(before)))
 	return output, err
 }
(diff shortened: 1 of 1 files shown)


harrywhite4:
Also seems that the problem only occurs in recent versions of docker-compose. I was using 1.24.1

PR #194, opened 2019-11-22
Update README.md

Olshansk:
- Link to an issue with instructions on how to copy-paste text on a Mac

* Not adding all the instructions directly into the README to avoid bloat


dawidd6:
Thanks!

PR #199, opened 2019-12-22
only use curl in install_update_linux.sh

### LINKED ISSUE #198: install_update_linux.sh: use either wget or curl, not both
When using that script in a CI chain, inside a minimalist container, it is nice not to have to install both wget and curl

gggeek:
Fixes #198


dawidd6:
Thanks.

PR #247, opened 2020-12-12
Exec shell: separate menu item and auto detect shell: bash or sh

stokito:


mjarkk:
I'm not supper active in this project so I have some questions before I take any actions.  

- i suspect that the host os and shell have no effect on this, is this correct?
- does this work on all containers? What if passwd doesn't exist or can't be opened?

stokito:
1. yes, the shell detection performed inside of container
2. After a lot of tests I found that only ClearLinux does't work. But the distro is very new and not popular yet. Their users (if any) may configure a custom command if they are really want. Meanwhile I hope that docker cli in a year or two will implement the `docker shell` command.

mjarkk:
Hmmm so i cannot get myself to agree with using inline shell commands like this: `docker exec -it {{ .Container.ID }} /bin/sh -c 'eval $(grep ^$(id -un): /etc/passwd | cut -d : -f 7-)'`  
It should work just fine but it just feels hackery to be honest.  

How about we do something else:  
We add a new template variable named `{{ .PreferredShell }}` that contains `/bin/bash` or `/bin/zsh` or `/bin/sh` ...  
Any time we execute a config command that has access to `{{ .Container.ID }}` *(a running container id)* we also check run this code:  
- Run: `docker exec -it {{ .Container.ID }} cat /etc/shells`
- If the output is successful we rank the output and pick the highest option like option one would be zsh, 2 = bash, 3 = sh
- If something failed anywhere we default to `/bin/sh`

stokito:
well... I sent PRs with the same cmd line to few others docker clients and it was merged. So it was several times reviewed and tested. It looks hacky, I aggrege, but anybody interested can easily debug this or just read a commit comment.
In fact the basic usage scenario is just to login into bash on Ubuntu based containers. There is no need to make some advanced strategy.
I think we shouldn't overcomplicate here but instead we may send PR to docker client with the new `docker shell` command

mjarkk:
Hmmm i'm still not completely for this change but knowing other clients also approve to this change and your word on it works in almost all containers gives me enough reasons to approve this change.
I'll merge this :)

mjarkk:
I've published a new release [v0.11](https://github.com/jesseduffield/lazydocker/releases/tag/v0.11)

PR #248, opened 2020-12-12
Open in browser

stokito:
Most containers expose some http port. We can open it in browser.
For simplicity we can just open first container's published port.

The same feature already exists in DockerStation

### DIFF
--- a/pkg/gui/containers_panel.go
+++ b/pkg/gui/containers_panel.go
@@ -571,3 +571,26 @@ func (gui *Gui) handleContainersBulkCommand(g *gocui.Gui, v *gocui.View) error {
 
 	return gui.createBulkCommandMenu(bulkCommands, commandObject)
 }
+
+// Open first port in browser
+func (gui *Gui) handleContainersOpenInBrowserCommand(g *gocui.Gui, v *gocui.View) error {
+	container, err := gui.getSelectedContainer()
+	if err != nil {
+		return nil
+	}
+	// skip if no any ports
+	if len(container.Container.Ports) == 0 {
+		return nil
+	}
+	// skip if the first port is not published
+	port := container.Container.Ports[0]
+	if port.IP == "" {
+		return nil
+	}
+	ip := port.IP
+	if ip == "0.0.0.0" {
+		ip = "localhost"
+	}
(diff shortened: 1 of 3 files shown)


pkg/gui/containers_panel.go
```diff
@@ -571,3 +571,26 @@ func (gui *Gui) handleContainersBulkCommand(g *gocui.Gui, v *gocui.View) error {
...
+	if err != nil {
+		return nil
+	}
+	// skip if no any ports
+	if len(container.Container.Ports) == 0 {
+		return nil
+	}
+	// skip if the first port is not published or it's a UDP port
+	port := container.Container.Ports[0]
+	if port.IP == "" || port.Type != "tcp" {
```
nurlansu, line 587: Just wondering, cant't we directly check if it's an UDP port?

```suggestion
	if port.IP == "" || port.Type == "udp" {
```

That would make more sense to the comment above.

stokito: but what if (one day) instead of tcp/udp there will be some another proto? This line is simple and fine, if you don't mind I'll keep it as is

mjarkk: the new HTTP/3 runs on top of QUIC and that's running on top of UDP so it's probably best to remove this type check.  
This protocol is also already used in major websites like google so it's not anymore some proposal that someday will be used.

mjarkk, line 587: the new HTTP/3 runs on top of QUIC and that's running on top of UDP so it's probably best to remove this type check.
This protocol is also already used in major websites like google so it's not anymore some proposal that someday will be used.

mjarkk (changes requested):

pkg/gui/containers_panel.go
stokito: Ok, then we may remove the check

stokito: I removed the check, please merge

mjarkk (approved):
LGTM

stokito:
@mjarkk while you are here please take a look on another my PR https://github.com/jesseduffield/lazydocker/pull/247

PR #252, opened 2021-03-26
Open service in browser

jtraub:
I really liked new features in 0.12 but unfortunately [Open in browser](#248) doesn't work in services panel.

This PR merely creates the same function for services panel. I took a quick look at `pkg/gui/service_panel.go` and `pkg/gui/containers_panel.go` and used copy-paste with slight adoption like I saw in other functions. Sorry if it was bad approach but I have no prior Go experience

### DIFF
--- a/pkg/gui/services_panel.go
+++ b/pkg/gui/services_panel.go
@@ -421,19 +421,5 @@ func (gui *Gui) handleServicesOpenInBrowserCommand(g *gocui.Gui, v *gocui.View) 
 		return gui.createErrorPanel(gui.g, gui.Tr.NoContainers)
 	}
 
-	// skip if no any ports
-	if len(container.Container.Ports) == 0 {
-		return nil
-	}
-	// skip if the first port is not published
-	port := container.Container.Ports[0]
-	if port.IP == "" {
-		return nil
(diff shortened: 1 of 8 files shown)


pkg/gui/services_panel.go
```diff
@@ -409,3 +409,31 @@ func (gui *Gui) handleServicesBulkCommand(g *gocui.Gui, v *gocui.View) error {
...
+	port := container.Container.Ports[0]
+	if port.IP == "" {
+		return nil
+	}
+	ip := port.IP
+	if ip == "0.0.0.0" {
+		ip = "localhost"
+	}
+	link := fmt.Sprintf("http://%s:%d/", ip, port.PublicPort)
+	return gui.OSCommand.OpenLink(link)
```
mjarkk, line 438: This seems to be the same code as in `container_panel.go` > `handleContainersOpenInBrowserCommand`  
Can you abstract this code so we don't have duplicated code?

mjarkk (changes requested):
One small issue.

jesseduffield:
Given this PR was a little stale I've taken the liberty of doing the refactoring :)

Issue #228, opened 2020-08-21
DOCKER_HOST variable causes crash


avkonst:
**Describe the bug**
If DOCKER_HOST environment variable is set, lazydocker crashes on start

**To Reproduce**

```
> export DOCKER_HOST=10.39.240.1:2375
> lazydocker 
2020/08/21 02:08:57 An error occurred! Please create an issue at https://github.com/jesseduffield/lazydocker/issues

*errors.errorString unable to parse docker host `10.39.240.1:2375`
/home/circleci/project/main.go:88 (0x8a7c99)
/usr/local/go/src/runtime/proc.go:200 (0x42d79c)
/usr/local/go/src/runtime/asm_amd64.s:1337 (0x4585b1)

> export DOCKER_HOST=10.39.240.1
> lazydocker 
2020/08/21 02:09:14 An error occurred! Please create an issue at https://github.com/jesseduffield/lazydocker/issues

*errors.errorString unable to parse docker host `10.39.240.1`
/home/circleci/project/main.go:88 (0x8a7c99)
/usr/local/go/src/runtime/proc.go:200 (0x42d79c)
/usr/local/go/src/runtime/asm_amd64.s:1337 (0x4585b1)
```

**Expected behaviour**
The tool connects to the target docker host

**Desktop (please complete the following information):**
 - Ubuntu 18.04
 - Lazydocker Version: 0.9

tylerstillwater:
Hit this also. It would be amazing to have this fixed.

tylerstillwater:
Editing the config to change the `docker-compose` command also doesn't seem to work:

```yaml
commandTemplates:
  dockerCompose: "docker-compose --context data"
```

tylerstillwater:
Turns out `ssh://` doesn't work for `DOCKER_HOST`. It causes lazydocker to crash.

I was able to work around this issue by creating an SSH tunnel to the remote docker socket like so:

`ssh -L /path/to/your/home/directory/or/something/docker-remote.sock:/var/run/docker.sock remote-ssh-host -fN`

Then setting the `DOCKER_HOST=unix:///path/to/your/home/directory/or/something/docker-remote.sock`

Issue #213, opened 2020-04-01
Can not use DOCKER_HOST with ssh and no -H command line option


bulletmark:
Docker allows users to set DOCKER_HOST using ssh but lazydocker fails. E.g.
```sh
pc:~ export DOCKER_HOST="ssh://vb1" 
pc:~ docker ps
CONTAINER ID        IMAGE               COMMAND                  CREATED             STATUS              PORTS                    NAMES
ee51f3403535        keepalived:latest   "/etc/keepalived/sta…"   19 minutes ago      Up 15 minutes                                keepalived
32f8de6c3b34        portainer/agent     "./agent"                19 minutes ago      Up 15 minutes       0.0.0.0:9001->9001/tcp   portainer
b38f73b4f61f        smsif2:latest       "python -u smsif2"       19 minutes ago      Up 15 minutes       0.0.0.0:8010->8010/tcp   smsif2

pc:~ lazydocker
2020/04/01 16:22:27 An error occurred! Please create an issue at https://github.com/jesseduffield/lazydocker/issues

*errors.withStack error during connect: Get "http://vb1/v1.25/images/json": dial tcp: lookup vb1: no such host
main.go:88 (0x8ba8e8)
/usr/lib/go/src/runtime/proc.go:203 (0x436fd2)
	main: fn()
/usr/lib/go/src/runtime/asm_amd64.s:1373 (0x464471)
	goexit: BYTE	$0x90	// NOP
```

Also, docker and most other docker tool provide `-H` to use instead of having to set DOCKER_HOST. Ideally lazydocker would add this.
```sh
pc:~ unset DOCKER_HOST
pc:~ docker -H="ssh://vb1" ps
CONTAINER ID        IMAGE               COMMAND                  CREATED             STATUS              PORTS                    NAMES
ee51f3403535        keepalived:latest   "/etc/keepalived/sta…"   22 minutes ago      Up 18 minutes                                keepalived
32f8de6c3b34        portainer/agent     "./agent"                22 minutes ago      Up 18 minutes       0.0.0.0:9001->9001/tcp   portainer
b38f73b4f61f        smsif2:latest       "python -u smsif2"       22 minutes ago      Up 18 minutes       0.0.0.0:8010->8010/tcp   smsif2

pc:~ lazydocker -H="ssh://vb1" 
lazydocker - The lazier way to manage everything docker
https://github.com/jesseduffield/lazydocker

  Flags: 
       --version   Displays the program version string.
[...]

PR #268, opened 2021-09-05
feat: support `ssh://` scheme for docker client host

### LINKED ISSUE #244: using lzd for connecting to remote docker host using ssh
ld(lazy docker) works fine with my localist docker daemon on Mac. when I am trying ld for connecting to remote docket host via SSH (18.09), it is throwing the following error.

=====
❯ ld
2020/11/16 22:46:00 An error occurred! Please create an issue at https://github.com/jesseduffield/lazydocker/issues

*errors.withStack error during connect: Get http://root%40172.xx.xx.xx/v1.25/images/json: dial tcp: lookup root@172.22.17.24: no such host
/home/circleci/project/main.go:88 (0x14a9449)
/usr/local/go/src/runtime/proc.go:200 (0x102d1cc)
/usr/local/go/src/runtime/asm_amd64.s:1337 (0x1057ed1)
░▒▓    ~                                                                                                                                                                                    1 ✘ ▓▒░
========

my env variable for DOCKER_HOST is set and also I am able to use the std docker cli from Mac to connect to remote daemon.
like : docker ps, docker service ls. It works. Is there a way I can enable ld for remote daemon connect with ssh.?

ps : screenshot above from terminal.

cmoog:
When the lazydocker `DOCKER_HOST`
environment variable is of scheme
`ssh://`, lazydocker will first open
an SSH tunnel to the specified host,
forwarding the docker.sock back to a local tmp file.

Then, lazydocker will use the tmp file
as the docker client socket.

Example usage: `DOCKER_HOST=ssh://my-server lazydocker`

Fixes #244, #228, 255


pkg/commands/docker.go
```diff
@@ -66,9 +72,113 @@ func (c *DockerCommand) NewCommandObject(obj CommandObject) CommandObject {
...
+		return err
+	}
+	defer conn.Close()
+	return nil
+}
+
+func tunnelSSH(ctx context.Context, host, localSocket string) error {
+	cmd := exec.CommandContext(ctx, "ssh", "-L", localSocket+":/var/run/docker.sock", host, "-N")
+	// when lazydocker exits, kill the SSH tunnel
+	cmd.SysProcAttr = &syscall.SysProcAttr{
```
jesseduffield, line 157: I'm not sure that we actually need to do this: when the parent process is killed, any children processes should also be killed.

jesseduffield (changes requested):
Sorry for taking so long: great work! Just two comments

pkg/commands/docker.go
cmoog: After removing this, I'm seeing the tunnel process leak of macOS, seen by the following command after opening and closing `lazydocker` with `<control-c>`.
```console
ps -ax | grep sshtunnel
```

jesseduffield: damn. Do we know if, with the SysProcAttr line still here, the child process is killed upon quitting via 'q'? If not, I wonder if we should have a pointer to this cmd living at the top level (against the gui or the app struct) which then gets killed explicitly upon quitting the program

cmoog: I just pushed a change that improves the correctness, and has behaved properly during my manual testing of each exit method (CTR-C, q).

https://github.com/jesseduffield/lazydocker/pull/268/commits/892fc090b62bbd9f11156abc5da8201cfe540a1b

cmoog:
Just discovered a strange issue when testing with a biometric protected ssh key where SSH authentication is continually requested many many times before the TUI is populated. Looking for a fix, but this certainly blocks merging.

cmoog:
Aha! I didn't realize the docker CLI was invoked [directly as well](https://github.com/cmoog/lazydocker/blob/892fc090b62bbd9f11156abc5da8201cfe540a1b/pkg/commands/docker.go#L237). The latest [commit](https://github.com/jesseduffield/lazydocker/pull/268/commits/0bd4168dc9fbb21cf27f4275d924af4a315a43c9) ensures that only 1 SSH connection is requested during the lazydocker session.

cmoog:
@jesseduffield bump

pkg/commands/docker.go
```diff
@@ -67,9 +73,134 @@ func (c *DockerCommand) NewCommandObject(obj CommandObject) CommandObject {
...
+	if err != nil {
+		return err
+	}
+	defer conn.Close()
+	return nil
+}
+
+// CloseDockerSocketConnection kills the docker socket SSH forwarding process, if it exists.
+//
+// If will exist when DOCKER_HOST has the protocol scheme `ssh://`.
```
jesseduffield, line 165:
```suggestion
// It will exist when DOCKER_HOST has the protocol scheme `ssh://`.
```

pkg/commands/docker.go
```diff
@@ -67,9 +73,134 @@ func (c *DockerCommand) NewCommandObject(obj CommandObject) CommandObject {
...
+func CloseDockerSocketConnection() {
+	if dockerSSHConnection != nil {
+		syscall.Kill(-dockerSSHConnection.Process.Pid, syscall.SIGKILL)
+	}
+}
+
+// dockerSSHConnection holds package-level state for the last-opened SSH tunnel to a remote docker socket.
+var dockerSSHConnection *exec.Cmd
```
jesseduffield, line 173: What are your thoughts on having this be a field on a struct that we modify, as opposed to a package-level variable? Reason being is that in my experience all package level variables inevitably cause problems for me. We could potentially store this on our App struct, a DockerCommand struct, or some new struct that contains all of these newly added methods.

jesseduffield (changes requested):
Couple more things, otherwise code all looks great

pkg/commands/docker.go
cmoog: I completely agree that package-level state is not a good solution here. I just pushed a change that abstracts the docker socket cleanup under the chain of `Close` methods... your call whether this is in fact cleaner https://github.com/jesseduffield/lazydocker/pull/268/commits/afe3d23cbd0345dfe786f4efa29283338dff5e79

jesseduffield:
@cmoog as part of pulling this down to test I made a couple changes and took the liberty of adding a test. I've moved the newly added functions into a new struct (and into a new ssh package for the sake of encapsulation). I've privat-ised some of the newly introduced types just because they're only used in the ssh package. For the sake of testing I've pulled all the methods that talk to the OS into a dependencies struct which is then passed mocks in the test.

Given the code you've added to the PR, I suspect you're a seasoned Go writer, so let me know if any of the changes I've made seem odd to you and whether you would tackle anything differently

pkg/commands/ssh/ssh.go
```diff
@@ -0,0 +1,156 @@
...
+	localSocket := path.Join(socketDir, "dockerhost.sock")
+
+	cmd, err := self.tunnelSSH(ctx, remoteHost, localSocket)
+	if err != nil {
+		return nil, fmt.Errorf("tunnel docker host over ssh: %w", err)
+	}
+
+	// set a reasonable timeout, then wait for the socket to dial successfully
+	// before attempting to create a new docker client
+	const socketTunnelTimeout = 8 * time.Second
```
cmoog, line 100: Maybe we should make this 15/20/30 seconds?

cmoog (commented):
Looks great to me @jesseduffield. Everything works as expected from my manual testing. Thanks for cleaning up the code and adding some proper test coverage.

jesseduffield (approved):
Looks good to me. Happy to merge when you are

cmoog:
LGMT! 🚢

n-bilalis:
Should this work on Windows too?

I'm getting the following error:

> tunnel ssh docker host: create ssh tunnel tmp file: CreateFile /tmp: The system cannot find the file specified.

mark2185:
@n-bilalis seems like an oversight. Could you please try building [this branch](https://github.com/mark2185/lazydocker/tree/fix/ssh-scheme-windows) from source?

n-bilalis:
> @n-bilalis seems like an oversight. Could you please try building [this branch](https://github.com/mark2185/lazydocker/tree/fix/ssh-scheme-windows) from source?

Thank you for the quick response!

I built and run it and now I get a different error:

> tunnel ssh docker host: ssh tunneled socket never became available: context deadline exceeded

It's probably related to this issue #213.

mark2185:
@n-bilalis could you pull and try again? I tried setting the context timeout to `5s` which could be enough.

n-bilalis:
Same error, unfortunately.

mark2185:
Eh, it was a longshot anyway. 

Sorry, don't have any more tricks up my sleeve without rolling them up.

PR #272, opened 2021-10-26
Sort containers by state 

### LINKED ISSUE #224: Sort containers by state
I have a lot of containers, and every time I have scroll to find the running containers, I spent time and is frustated.
I would like to have an option to (o)sort and if is possible have a default order: 1-(r)running, 2-(e)exited, 3-(c)created.

lpessoa:
Fixes #224 by enabling sorting container by state using 'o' menu shortcut.

399633D5-32DA-4590-9A41-82204D02AD83

- [X] Functionality
- [x] Docs
- [x] CI build


jesseduffield:
Sorry for my delayed review!
Thinking out loud: in the abstract, it seems like we should have our container view's behaviour consistent with our services view given that they both basically list containers which could be running/exited/etc. But with services, you're never going to have that many, and sorting by whether they'res a container at all seems pretty sufficient because then you don't need to look so hard for any given service given they're always in the same order (so long as there's a container there).

With containers on the other hand, it seems like basically everybody has to deal with a bunch of exited containers that can make it hard to spot the ones that aren't exited. For this reason I'm actually thinking we should just make sorting of containers, as contained in this PR, the default. Rather than have a keybinding for sorting, I would add a config option that allows the user to specify that they want the old behaviour of sorting by name only. What do you think?

lpessoa:
Sorry for the late reply too 😄 .

Sounds good to me. Will make it default and have a config option to allow user defined sorting 😉

lpessoa:
@jesseduffield done.. included some tests too

pkg/commands/docker.go
```diff
@@ -274,6 +275,22 @@ func (c *DockerCommand) filterOutExited(containers []*Container) []*Container {
 	return toReturn
 }
 
+// sortedContainers returns containers sorted by state if c.SortContainersByState is true (follows 1- running, 2- exited, 3- created)
+// and sorted by name if c.SortContainersByState is false
+func (c *DockerCommand) sortedContainers(containers []*Container) []*Container {
+	if !c.Config.UserConfig.Gui.SortContainersByName {
+		sort.Slice(containers, func(i, j int) bool {
+			states := map[string]int{
```
jesseduffield, line 283: Given that this is effectively just a constant, we should pull it out of the loop to spare ourselves the unnecessary memory allocations

jesseduffield (changes requested):
Nice work! Just one comment

pkg/commands/docker.go
lpessoa: Done 😄

jesseduffield:
@lpessoa just did a local test and fixed a bug where it crashed upon startup due to a malformed struct field tag. I also noticed that when `sortContainersByName` is true in my config the containers aren't actually sorted by name. We should either sort them by name in the code or change the config key to be more accurate

lpessoa:
> @lpessoa just did a local test and fixed a bug where it crashed upon startup due to a malformed struct field tag. I also noticed that when `sortContainersByName` is true in my config the containers aren't actually sorted by name. We should either sort them by name in the code or change the config key to be more accurate

I have developed the feature to default to the legacy behaviour when the setting is true but forgot to name the var accordingly 😞 (i was thinking on sort by name when developing it).
Do you want to keep the old behavior and change the setting to something like `LegacySortContainers`?

jesseduffield:
@lpessoa yeah LegacySortContainers works for me

lpessoa:
@jesseduffield `LegacySortContainers`set for legacy behaviour. Updated the new behaviour to sort by state **and** name

jesseduffield:
Nice work!

MartinCura:
Just as an FYI, i'm happy i found this PR because i find it very annoying that when i start a container the selected container is changed as they're resorted. Thus i'm now using the undocumented `legacySortContainers`. Another fix would be that the selection is persisted on container state change! (btw thank you for lazydocker, only just found about it).

jesseduffield:
@MartinCura could you raise an issue for that?

PR #280, opened 2021-12-02
Fix AUR link in docs

conorbros:


mjarkk:
Hmm maybe we should simplify the arch section because most people who use arch probably know how to install a package from the AUR.

```md
### Arch Linux AUR

You can install [lazydocker from the AUR](https://aur.archlinux.org/packages/lazydocker)
```

Also the arch section is now a bit confusing as it gives the impression there is a different package for the development version while the url points to the same package as installed in the example above it

conorbros:
Fair enough, I just wanted something to grab and copy paste.

mjarkk:
Maybe we can include the yay example as it's a commonly used package manager for arch

conorbros:
Sounds good to me. I think most Arch users would be happy with the yay example

mjarkk:
Thanks

Issue #126, opened 2019-07-11
Key Binding for Panel Nav?


andyl:
Is there a keybinding to navigate between the `Project / Services / Containers / Images / Volume` panels?  (I see that `enter` takes you to the main panel...)

andyl:
Oh just found it - left and right arrow key...

PR #288, opened 2021-12-28
Add panel navigation to keybindings description

lknop:
@jesseduffield I spent last 15 minutes looking for this information, I tried up/down arrows, tabbing, page up/page down, I looked through the documentation, watched the youtube video, googled, read a few reddit threads, youtube comments, and did multiple searches of the issues. I even remember installing lazydocker a like a year ago and removing it after not finding the panel navigation. This missing info was mentioned in #126 but it was closed without any change.

### DIFF
--- a/docs/keybindings/Keybindings_en.md
+++ b/docs/keybindings/Keybindings_en.md
@@ -72,5 +72,6 @@
 
 <pre>
   <kbd>esc</kbd>: return
-  <kbd>left/right</kbd> panel navigation
+  <kbd>left</kbd>: previous panel
+  <kbd>right</kbd>: next panel
 </pre>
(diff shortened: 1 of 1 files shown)


jesseduffield:
Sorry about your troubles, we really should just have tab/shift-tab as valid keybindings now that our gocui dependency supports it (but I'm not sure how easy it is to bump the version: I've made quite a few changes for the sake of my other project Lazygit in the gocui repo).

As for this change, the file you've updated is auto-generated by `scripts/generate_cheatsheet.go`. We'll need to either make a change to that script, or to `pkg/gui/keybindings.go`, so that the keybindings appear

jesseduffield:
closing due to stalenes

PR #291, opened 2022-01-08
Update `go install` instructions

dezren39:
https://go.dev/doc/go-get-install-deprecation


jesseduffield:
thanks for making this!

Issue #293, opened 2022-01-11
[Feature Request] Creating a seperate context to display just environment variables in the container panel


glendsoza:
**Is your feature request related to a problem? Please describe.**
Usually, applications running inside the docker container are configured using environment variables, hence I usually find myself checking environment variables set in the container 

Currently, if I want to see environment variables for a container I need to go to the `Config` section and looks for `Env` section

It would be great if there is a separate tab where I can just view the Environment variables 

**Describe the solution you'd like**

Something like this

![with_env](https://user-images.githubusercontent.com/95934259/148991380-17615692-29fd-499f-bd8a-e31e43407428.png)

**Describe alternatives you've considered**

**Additional context**

With some code changes, I was able to implement this feature (as shown in the screenshot above), if this gets approved would be glad to create the PR!

jesseduffield:
I'm on board with this idea. Happy to review a PR :)

glendsoza:
Here is the PR https://github.com/jesseduffield/lazydocker/pull/295, please let me know your thoughts on this

PR #295, opened 2022-01-12
Created a seperate context to display just environment variables

glendsoza:
This PR resolves https://github.com/jesseduffield/lazydocker/issues/293

There is one issue with the implementation, in order to show environment variables in a presentable way they are padded, this might result in some values getting trimmed i.e they won't be displayed fully. I was thinking to split either key or value if they are too long but this will look messy hence decided to go with the current implementation.
 
Any suggestion to overcome this issue is welcome


pkg/gui/containers_panel.go
```diff
@@ -90,6 +94,37 @@ func (gui *Gui) handleContainerSelect(g *gocui.Gui, v *gocui.View) error {
...
 }
 
+func (gui *Gui) renderContainerEnv(container *commands.Container) error {
+	mainView := gui.getMainView()
+	mainView.Autoscroll = false
+	mainView.Wrap = gui.Config.UserConfig.Gui.WrapMainPanel
+	outputList := []string{""}
+	envMapping := map[string]string{}
+	padding := 0
+	if len(container.Details.Config.Env) > 0 {
```
jesseduffield, line 104: We can use the `utils.RenderTable` function here to handle the padding. It takes a slice of string slices, so we can pass it e.g. `[['MY_ENV', 'MY_ENV_VALUE'], ['MY_OTHER_ENV', 'MY__OTHER_ENV_VALUE']]` and it'll handle the padding. I think you can colour the strings beforehand but if not you can use `utils.RenderList` which has some example uses lying around as well

jesseduffield (changes requested):
One thing :)

pkg/gui/containers_panel.go
glendsoza: Thanks for your time
I have made changes to use `utils.RenderTable` function with coloring strings beforehand
I have also made changes to `utils.RenderTable` method to not to panic when an empty array is passed to it
Please review the changes

pkg/gui/containers_panel.go
```diff
@@ -18,11 +18,11 @@ import (
 // list panel functions
 
 func (gui *Gui) getContainerContexts() []string {
-	return []string{"logs", "stats", "config", "top"}
+	return []string{"logs", "stats", "env", "config", "top"}
```
jesseduffield, line 21: could we also add this for the services panel?

pkg/gui/containers_panel.go
```diff
@@ -90,6 +94,37 @@ func (gui *Gui) handleContainerSelect(g *gocui.Gui, v *gocui.View) error {
...
+	mainView.Wrap = gui.Config.UserConfig.Gui.WrapMainPanel
+	envVariablesList := [][]string{}
+	if len(container.Details.Config.Env) > 0 {
+		for _, env := range container.Details.Config.Env {
+			splitEnv := strings.Split(env, "=")
+			// if the value has = in it, lets say export "test=foo=bar" then split will result in the following
+			// {"test", "foo","bar"} hence join all the elements in the slice except the first one to get value
+			envVariablesList = append(envVariablesList,
+				[]string{
+					utils.ColoredString(splitEnv[0]+":", color.FgBlue),
```
jesseduffield, line 109: Blue is a little hard to see on some terminals, so I reserve it for things that aren't too important to see (like the `none` status of a service). Could we instead use green instead?

pkg/gui/containers_panel.go
```diff
@@ -90,6 +94,37 @@ func (gui *Gui) handleContainerSelect(g *gocui.Gui, v *gocui.View) error {
...
 }
 
+func (gui *Gui) renderContainerEnv(container *commands.Container) error {
+	mainView := gui.getMainView()
+	mainView.Autoscroll = false
+	mainView.Wrap = gui.Config.UserConfig.Gui.WrapMainPanel
+	envVariablesList := [][]string{}
+	if len(container.Details.Config.Env) > 0 {
+		for _, env := range container.Details.Config.Env {
+			splitEnv := strings.Split(env, "=")
```
jesseduffield, line 104: we can instead use `strings.SplitN(str, "=", 2)` here to only split on the first equals sign

pkg/gui/containers_panel.go
```diff
@@ -90,6 +94,37 @@ func (gui *Gui) handleContainerSelect(g *gocui.Gui, v *gocui.View) error {
...
+			// if the value has = in it, lets say export "test=foo=bar" then split will result in the following
+			// {"test", "foo","bar"} hence join all the elements in the slice except the first one to get value
+			envVariablesList = append(envVariablesList,
+				[]string{
+					utils.ColoredString(splitEnv[0]+":", color.FgBlue),
+					utils.ColoredString(strings.Join(splitEnv[1:], "="), color.FgYellow),
+				})
+		}
+	} else {
+		envVariablesList = append(envVariablesList, []string{"Nothing to display"})
```
jesseduffield, line 114: can we internationalise these strings? See `pkg/i18n/english.go` for examples. No need to translate into other languages: just having english versions is sufficient

pkg/gui/containers_panel.go
```diff
@@ -90,6 +94,37 @@ func (gui *Gui) handleContainerSelect(g *gocui.Gui, v *gocui.View) error {
...
+	}
+	renderedTable, err := utils.RenderTable(envVariablesList)
+	// in case of some error
```
jesseduffield, line 117: no need for this comment because the code is fairly self-explanatory.

pkg/gui/containers_panel.go
```diff
@@ -90,6 +94,37 @@ func (gui *Gui) handleContainerSelect(g *gocui.Gui, v *gocui.View) error {
...
+	if err != nil {
+		// log the error
```
jesseduffield, line 119: likewise here. I only add comments where I need to explain the something that isn't well communicated by the code itself

pkg/gui/containers_panel.go
jesseduffield: Also seems like there's no need to go through the table approach here. We could just set `renderedTable` directly

pkg/utils/utils.go
```diff
@@ -165,6 +165,10 @@ func renderDisplayableList(items []Displayable, config RenderListConfig) (string
 
 // RenderTable takes an array of string arrays and returns a table containing the values
 func RenderTable(stringArrays [][]string) (string, error) {
+	// if empty array is given then getPadWidths will panic hence check to make sure array is not empty
+	if len(stringArrays) == 0 {
+		return "", errors.New("RenderTable given an empty array")
```
jesseduffield, line 170: I don't actually think this is worthy of an error. I would instead simply return an empty string and a nil error.

jesseduffield (changes requested):
Great feature! I've left some feedback

pkg/gui/containers_panel.go
glendsoza: Added the `container-env` section for services panel

pkg/gui/containers_panel.go
glendsoza: Changed the color to blue

pkg/gui/containers_panel.go
glendsoza: Used the `strings.SplitN(str, "=", 2)`

pkg/gui/containers_panel.go
glendsoza: Removed any unnecessary comments

pkg/utils/utils.go
glendsoza: Returning an empty string and nil error if `len(stringArrays) == 0`

pkg/gui/containers_panel.go
glendsoza: internationalised the strings and also altered the logic to set `renderedTable`

pkg/i18n/turkish.go
```diff
@@ -63,11 +63,15 @@ func turkishSet() TranslationSet {
...
 		ConfigTitle:               "Ayarlar",
+		EnvTitle:                  "Env",
 		DockerComposeConfigTitle:  "Docker-Compose Ayar",
 		TopTitle:                  "Top",
 		StatsTitle:                "Durumlar",
 		CreditsTitle:              "Hakkinda",
 		ContainerConfigTitle:      "Konteyner Ayar",
+		ContainerEnvTitle:         "Konteyner Env",
+		NothingToDisplay:          "Nothing to display",
+		CannotDisplayEnvVairables:  "Something went wrong while displaying environment variables",
```
jesseduffield, line 74:
```suggestion
		CannotDisplayEnvVariables:  "Something went wrong while displaying environment variables",
```

pkg/utils/utils.go
```diff
@@ -165,6 +165,10 @@ func renderDisplayableList(items []Displayable, config RenderListConfig) (string
 
 // RenderTable takes an array of string arrays and returns a table containing the values
 func RenderTable(stringArrays [][]string) (string, error) {
+	// if empty array is given then getPadWidths will panic hence return an empty string
```
jesseduffield, line 168: I'd lose this comment too because getPadWidths panicking isn't really the reason we're returning an empty string: it's more that we know we don't need to render anything and I think the code is self-explanatory in that case

jesseduffield (changes requested):
two more things, then we should be good!

glendsoza:
@jesseduffield Thank you for taking the time to review, I have made changes as per your last review

jesseduffield:
Looks like there's a couple of formatting issues to fix up @glendsoza https://app.circleci.com/pipelines/github/jesseduffield/lazydocker/529/workflows/4be1cc83-2c61-4d8b-b089-2aaf3650d980/jobs/682

glendsoza:
@jesseduffield Fixed the issues with formatting

jesseduffield:
Merged! I'll need to go fix up that windows-specific test some time, but once I've done that I'll cut a new release. Thanks for making this, it's gonna be super handy

PR #297, opened 2022-01-14
Correct previous view behaviour

### LINKED ISSUE #298: Correct previous view behaviour
First of all thanks a lot for your work <3
And an excuse to dig deep(er) into go! :)

**Return from shell**

Let's say we start at the containers view
// action: current view - previous views stack (from the top)
init: containers - []
switch to: images - [containers]
switch to: containers - [images]
shell and back: images - []
But I think it makes much more sense to stay at containers from where we fired a shell.

Function of interest: `gui.RunWithSubprocesses`
Currently we push to the stack, then run a command, then pop what we pushed, which doesn't seem to change anything. I propose to change the order of push and pop, so that containers view got properly restored on return.

**Return from context menu**

Let's say we start at the containers view
// action: current view - previous views stack
init: containers - []
switch to: images - [containers]
switch to: menu - [images containers]
switch to: bulk - [containers]
cancel: containers - []
But again probably we should've returned to images from where the menu got invoked.

Functions of interest:
`gui.returnFocus`
Currently we push the previous view only if it's not a popup but pop regardless. I propose we do the pop in only if the current view is not a popup, why pop if we didn't push for it?
`gui.switchFocus`
Pop here is needed so that the views don't stack.

Tested on current master (12 Jan), ubuntu 18.04

Here's my stab at that: https://github.com/jesseduffield/lazydocker/pull/297

bravmi:
Resolves #298

### DIFF
--- a/pkg/gui/subprocess.go
+++ b/pkg/gui/subprocess.go
@@ -21,6 +21,8 @@ func (gui *Gui) RunWithSubprocesses() error {
 			if err == gocui.ErrQuit {
 				break
 			} else if err == gui.Errors.ErrSubProcess {
+				// pop here so we don't stack up view names
+				gui.popPreviousView()
 				// preparing the state for when we return
 				gui.pushPreviousView(gui.currentViewName())
 				// giving goEvery goroutines time to finish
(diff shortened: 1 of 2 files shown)


jesseduffield:
@bravmi this is looking good but it introduces a new problem. If you go to the images view, then press enter to go to the main view, then open a menu, then hit escape, you actually need to hit escape two more times to get back to a side view, and we land in the services view rather than the images view.

Not sure why that's happening but happy to help out some more if you need

jesseduffield (changes requested):
marking as changes requested for my own bookkeeping

jesseduffield:
Closing as this issue is fixed in https://github.com/jesseduffield/lazydocker/pull/372

Issue #260, opened 2021-06-14
LazyDocker does not show all images


MiroslavMikus:
**Describe the bug**
`docker images` shows more images than lazy-docker

**To Reproduce**
1. pull image from private repository
2. run `docker images`

**Expected behaviour**
lazy-docker should show all images from `docker images` list

**Screenshots**
![image](https://user-images.githubusercontent.com/29312172/121902887-08812780-cd28-11eb-8777-4b66fd04ca8e.png)

**Desktop (please complete the following information):**
 - OS: Windows Server 2019. **It works on on windows 10**
 - Lazydocker Version 0.12

dvjones89:
Hi @MiroslavMikus 👋  Forgive me if I'm telling you something you already know, but have you tried giving focus to the "images" pane within `lazydocker` and then pressing the "up" and "down" arrows on your keyboard? The pane is scrollable, though it's perhaps not immediately obvious from the user interface, hence my question.

MiroslavMikus:
@dvjones89 Hey, good point! Hmm... tried again with the same result:
![image](https://user-images.githubusercontent.com/29312172/122547962-0de4b780-d031-11eb-8418-e6631a180c12.png)

VProfirov:
Lazydocker doesn't refresh for new images - created after starting the lazydocker.

Can this be quickly fixed as it is for containers?

Thank you in advance! Great tool, thank you for which! :)

jesseduffield:
@VProfirov that should be an easy fix. Happy to accept a PR for this, and give any guidance needed

glendsoza:
Here is a PR which tries to add this functionality https://github.com/jesseduffield/lazydocker/pull/299

PR #299, opened 2022-01-15
Changes to auto refresh the images every 30s

glendsoza:
This PR tries to resolve https://github.com/jesseduffield/lazydocker/issues/260

A couple of Questions:

1. As per the PR I have hardcoded the auto-refresh interval to be 30s, should I make it configurable via config file like how it's done for docker refresh interval
2. Since images are getting refreshed on an interval the function `refreshSidePanels` does not do anything hence I have commented it out, should I remove that function and make changes to any code which calls it?

### DIFF
--- a/pkg/commands/docker.go
+++ b/pkg/commands/docker.go
@@ -15,6 +15,7 @@ import (
 
 	"github.com/acarl005/stripansi"
 	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/api/types/filters"
 	"github.com/docker/docker/client"
 	"github.com/imdario/mergo"
 	"github.com/jesseduffield/lazydocker/pkg/commands/ssh"
@@ -40,6 +41,7 @@ type DockerCommand struct {
 	ErrorChan              chan error
 	ContainerMutex         sync.Mutex
 	ServiceMutex           sync.Mutex
+	ImageMutex             sync.Mutex
 	Services               []*Service
 	Containers             []*Container
 	// DisplayContainers is the array of containers we will display in the containers panel. If Gui.ShowAllContainers is false, this will only be those containers which aren't based on a service. This reduces clutter and duplication in the UI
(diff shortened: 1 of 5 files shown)


pkg/gui/view_helpers.go
```diff
@@ -12,9 +12,9 @@ import (
 
 func (gui *Gui) refreshSidePanels(g *gocui.Gui) error {
 	// not refreshing containers and services here given that we do it every few milliseconds anyway
-	if err := gui.refreshImages(); err != nil {
-		return err
-	}
+	// if err := gui.refreshImages(); err != nil {
```
jesseduffield, line 15: I would keep this here. We call refreshSidePanels upon performing some action e.g. removing an image. We'd want that change to be reflected in the UI immediately.

pkg/gui/gui.go
```diff
@@ -246,6 +246,7 @@ func (gui *Gui) Run() error {
 		gui.goEvery(time.Millisecond*30, gui.reRenderMain)
 		gui.goEvery(dockerRefreshInterval, gui.refreshProject)
 		gui.goEvery(dockerRefreshInterval, gui.refreshContainersAndServices)
+		gui.goEvery(time.Millisecond*30000, gui.refreshImages)
```
jesseduffield, line 249: I would indeed make this configurable, defaulting to 30s

pkg/commands/docker.go
```diff
@@ -452,3 +454,53 @@ func (c *DockerCommand) DockerComposeConfig() string {
...
+	defer c.ImageMutex.Unlock()
+	images, err := c.Client.ImageList(context.Background(), types.ImageListOptions{})
+	if err != nil {
+		return err
+	}
+
+	ownImages := make([]*Image, len(images))
+
+	for i, image := range images {
+		// func (cli *Client) ImageHistory(ctx context.Context, imageID string) ([]image.HistoryResponseItem, error)
```
jesseduffield, line 470: we can remove this comment

jesseduffield (changes requested):
left some feedback :)

pkg/gui/gui.go
glendsoza: Made the refresh time to be configurable

pkg/gui/view_helpers.go
glendsoza: Method `refreshSidePanels` is only called in `loadNewDirectory` which in turn gets called only once 

In the current code base to refresh the image section after removing/pruning an image method `gui.refreshImages()` gets called 

Since now `gui.refreshImages()` gets called on a timer `refreshSidePanels` basically does not do anything

PR #302, opened 2022-01-18
Add key bindings to switch focus between views with number keys

### LINKED ISSUE #301: Allow navigating panes with the number keys
Whenever I use https://github.com/jesseduffield/lazygit, I navigate through the panes using the number keys (1-5) instead of mashing H or L. Whenever I use lazydocker, I immediately notice the lack of these hotkeys. Is there any reason why this shouldn't also be supported for lazydocker?

gekoke:
Resolves #301

### DIFF
--- a/pkg/gui/keybindings.go
+++ b/pkg/gui/keybindings.go
@@ -487,6 +487,11 @@ func (gui *Gui) GetInitialKeybindings() []*Binding {
 			{ViewName: viewName, Key: gocui.KeyArrowRight, Modifier: gocui.ModNone, Handler: gui.nextView},
 			{ViewName: viewName, Key: 'h', Modifier: gocui.ModNone, Handler: gui.previousView},
 			{ViewName: viewName, Key: 'l', Modifier: gocui.ModNone, Handler: gui.nextView},
+			{ViewName: viewName, Key: '1', Modifier: gocui.ModNone, Handler: gui.switchToProjectView},
+			{ViewName: viewName, Key: '2', Modifier: gocui.ModNone, Handler: gui.switchToServicesView},
+			{ViewName: viewName, Key: '3', Modifier: gocui.ModNone, Handler: gui.switchToContainersView},
+			{ViewName: viewName, Key: '4', Modifier: gocui.ModNone, Handler: gui.switchToImagesView},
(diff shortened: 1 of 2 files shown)


gekoke (commented):
Not a huge fan of this duplication, but it's the best I could come up with given the interface constraints

jesseduffield:
Nice work @gekoke  One thing we'll want to add here is to ignore pressing '2' if the services view is not shown. The services view is not shown when running lazydocker outside a docker compose project.

Also you'll need to rebase this onto master.

jesseduffield (changes requested):
marking as 'changes requested' for my own bookkeeping. Lemme know if you need any pointers :)

gekoke:
Ah, good catch. Trying out the hotkeys when there is no services view raises another question, though. Namely, that it feels a bit unnatural for the (at the moment, default) hotkeys to progress in the sequence 1, 3, 4, 5. We might want to switch to a view by index instead, but then the hotkey for switching to a particular view might differ depending on context.

Lazygit doesn't seem to have this problem, since, as far as I can tell, it displays all of its empty containers (e.g. when there is nothing in the stash).

I suppose what I'm describing are actually two separate pieces of functionality, and that the more natural option - for number keys, at least - is to switch by index.
 
 As a sidenote, is it bad practice to amend commits during a pull request? I want to keep the commits' messages sementically relevant, and the code bug-free, but I suppose that might detract from seeing the progression of a PR.

jesseduffield:
yeah going by index makes sense. The worst case scenario is somebody has muscle memory to press '4' to go to the images view but when not in a docker compose project that ends up taking them to the volumes view. But I suspect the typical user is not often jumping between docker compose and non-docker compose directories. And it would be weird to make it non-index based and say that if you happen to not use docker compose then you awkwardly go from 1 to 3 and 2 does nothing.

As for amending commits, I'm not really fussed when there's a small amount of code in the PR (as is the case here). I typically just re-read the full diff anyway.

jesseduffield:
@gekoke there's been some recent changes to master so this will need to be tweaked a bit (but not much). Are you still keen on continuing with this PR?

jesseduffield:
Closing due to staleness

PR #310, opened 2022-02-09
Fix the installed path written in the document to match the script

Tiryoh:
First of all, thanks for providing the awesome tool!
This PR fixes the installed path written in the document to match the script.

The description was added on PR https://github.com/jesseduffield/lazydocker/pull/209, and the script was updated on PR https://github.com/jesseduffield/lazydocker/pull/271, but the document was not updated.


jesseduffield:
Nice one

Issue #273, opened 2021-10-27
Warning: Calling bottle :unneeded is deprecated!


studious:
I'm just reporting this here. Same as lazygit

Cheers

den-is:
Please help brew to stop producing that annoying warning :((((

ocsw:
bump?

den-is:
@jesseduffield is it possible to fix that, that should be taking much time and is not touching actual application's logic and code

jesseduffield:
@den-is apologies I've made the grave error of maintaining two open source repos haha: lazygit certainly gets a lot more love.

As for this issue, it's a result of goreleaser using the deprecated bottling approach. That was fixed in goreleaser back in [october](https://github.com/goreleaser/goreleaser/issues/2595). We always use the latest goreleaser on release, however our CI is currently failing due to some code that windows does not support.

I've got a PR up to address this but will need help from people who use an SSH connection to docker to see if any child zombie processes are left after running lazydocker (or alternatively, if anybody can guide me on setting up docker via SSH on my mac I can test this out myself)

omaksi:
I've just tried installing and I'm actually getting an error for this, and it won't install from your tap.

```Error: Invalid formula: /usr/local/Homebrew/Library/Taps/jesseduffield/homebrew-lazydocker/lazydocker.rb
lazydocker: Calling bottle :unneeded is disabled! There is no replacement.
Please report this issue to the jesseduffield/lazydocker tap (not Homebrew/brew or Homebrew/core):
  /usr/local/Homebrew/Library/Taps/jesseduffield/homebrew-lazydocker/lazydocker.rb:9

Error: Cannot tap jesseduffield/lazydocker: invalid syntax in tap!
```

Installing from core works OK.

mark2185:
I have a `mac` lying around, what do I need to do? Just `brew install lazydocker`?

jesseduffield:
Nice. The core homebrew repo is yet to merge the lazydocker PR but you can test out the tap like so:
```
brew install jesseduffield/lazydocker/lazydocker
```

jesseduffield:
Woohoo! Thanks @mark2185 , let's close this off

PR #185, opened 2019-11-11
Fix #29: Windows Build

reicheltp:
On Windows we do not have `Setpgid` therefore `func (c *OSCommand) PrepareForChildren(cmd *exec.Cmd)` and `func (c *OSCommand) Kill(cmd *exec.Cmd) error` had to be replaced by windows specific implementations.

We use `kernel32.dll` to fetch the related processes to effectively kill all spawned child processes.

Follows an inspiration of: https://blog.csdn.net/fyxichen/article/details/51857864

Fixes #29

PR #313, opened 2022-02-22
fix CI

jesseduffield:
Currently our CI fails because windows does not support syscall.Kill, nor does it support SetPgid. This PR instead replaces it with a regular process kill. 

These two things were added in this commit: https://github.com/jesseduffield/lazydocker/commit/892fc090b62bbd9f11156abc5da8201cfe540a1b

But I'm not sure if they're necessary. I want to know whether the ssh command here ever creates its own child processes.

There was some discussion on this here: https://github.com/jesseduffield/lazydocker/pull/268/commits/4e201bb8f44342c3a5900cf286bbbc5343bc4da0#r757413230

I don't have an ssh setup on my own machine so it's tricky for me to test this. @cmoog can I enlist you to take another look at this?

If the PR in its current form has issues we'll need to find a way to do this which is compatible with windows.

See https://github.com/jesseduffield/lazydocker/issues/273 for discussion

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -60,20 +60,6 @@ jobs:
       - image: circleci/golang:1.17
     steps:
       - checkout
-      #- run:
-      #    name: Install snapcraft
-      #    command: |
-      #      ./.circleci/install_snapcraft.sh
-      #- run:
-      #    name: Login to snapcraft
-      #    # if you ever lose this, you can recreate it via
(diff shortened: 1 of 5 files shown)


jesseduffield:
Alright so I found some code that's specifically build around ensuring we can kill subprocesses  thanks to https://github.com/jesseduffield/lazydocker/pull/185. So this is good to go

PR #317, opened 2022-03-23
Add docker pause functionality

mark2185:
I'm not sure what to do about translations, though.

### DIFF
--- a/pkg/gui/containers_panel.go
+++ b/pkg/gui/containers_panel.go
@@ -423,13 +423,8 @@ func (gui *Gui) handleContainersRemoveMenu(g *gocui.Gui, v *gocui.View) error {
 	return gui.createMenu("", options, len(options), handleMenuPress)
 }
 
-func (gui *Gui) handleContainerPause(g *gocui.Gui, v *gocui.View) error {
-	container, err := gui.getSelectedContainer()
-	if err != nil {
-		return nil
-	}
-
-	return gui.WithWaitingStatus(gui.Tr.PausingStatus, func() error {
+func (gui *Gui) PauseContainer(container *commands.Container) error {
+	return gui.WithWaitingStatus(gui.Tr.PausingStatus, func() (err error) {
 		if container.Details.State.Paused {
 			err = container.Unpause()
 		} else {
@@ -442,6 +437,15 @@ func (gui *Gui) handleContainerPause(g *gocui.Gui, v *gocui.View) error {
 
 		return gui.refreshContainersAndServices()
 	})
+}
+
+func (gui *Gui) handleContainerPause(g *gocui.Gui, v *gocui.View) error {
+	container, err := gui.getSelectedContainer()
+	if err != nil {
+		return nil
+	}
+
+	return gui.PauseContainer(container)
 }
 
 func (gui *Gui) handleContainerStop(g *gocui.Gui, v *gocui.View) error {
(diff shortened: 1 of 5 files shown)


jesseduffield (approved):
LGTM

jesseduffield:
Actually I've just tested this out locally and looks like we're missing a loading state when pausing. Also, would we be able to add a similar keybinding for the services view so that it's consistent with the containers view? The services view appears when running lazydocker in a docker compose project

jesseduffield (changes requested):
changing to 'changes requested' for my own bookkeeping. Lemme know if you need any pointers :)

mark2185:
> would we be able to add a similar keybinding for the services view so that it's consistent with the containers view

We sure could! I'll also add the required test files for future development since I don't usually use `docker-compose`

The `loading state when pausing` is referred to something like "Pausing..."?

jesseduffield:
@mark2185 yep 'pausing' works (lowercase for consistency with other loading statuses)

mark2185:
I walked back out on adding necessary test files since the change was simple and I don't think I'll use `docker-compose` in the short term so I'd have to read up on it, and I don't really feel like it. Could you give it a spin?

pkg/gui/containers_panel.go
```diff
@@ -496,6 +496,27 @@ func (gui *Gui) handleContainersRemoveMenu(g *gocui.Gui, v *gocui.View) error {
 	return gui.createMenu("", options, len(options), handleMenuPress)
 }
 
+func (gui *Gui) handleContainerPause(g *gocui.Gui, v *gocui.View) error {
+	container, err := gui.getSelectedContainer()
+	if err != nil {
+		return nil
+	}
+
+	return gui.WithWaitingStatus(gui.Tr.PausingStatus, func() error {
```
jesseduffield, line 505: I'd extract this part out into its own function and call that from handleServicePause so that we're not duplicating the logic

pkg/gui/containers_panel.go
```diff
@@ -496,6 +496,27 @@ func (gui *Gui) handleContainersRemoveMenu(g *gocui.Gui, v *gocui.View) error {
...
+			err = container.Unpause()
+		} else {
+			err = container.Pause()
+		}
+
+		if err != nil {
+			return gui.createErrorPanel(gui.g, err.Error())
+		}
+
+		return gui.refreshContainersAndServices()
```
jesseduffield, line 516: I'd see how it goes without this line here. Lazydocker should automatically recognise that the container is paused and then refresh accordingly

jesseduffield (changes requested):
Looking good, just a couple things. Happy to test upon next review

pkg/gui/containers_panel.go
mark2185: It does not.

pkg/gui/containers_panel.go
mark2185: I've pushed it in a new commit so when you find the time you can easily test out the automatic recognizing of pausedness (I know it's not a word, but I do think it should be, haters be damned).

pkg/gui/containers_panel.go
jesseduffield: I've done some digging and this looks to be an actual bug in the docker package where we refetch container info after docker tells us there's been a change but the change doesn't come through. Crazy!

mark2185: Alright, now that we have the murderer, what are we going to do with it? The victim is apparently fine with `gui.refreshContainersAndServices()`

jesseduffield: Yeah let's just stick to calling `refreshContainersAndServices` and we can look into the deeper issue in another PR

jesseduffield:
nice work @mark2185

PR #328, opened 2022-05-10
reduce occurrences of 'unversioned' version

jesseduffield:
In go 1.18 the git version is by default added to the binary. So, if we don't have a git tag baked into the binary (e.g. because it was built by source), we're going to instead display the abbreviated commit SHA as the version. Unfortunately, 1.18 does not include the current git tag but the revision is still an improvement upon 'unversioned'.


pkg/gui/layout.go
```diff
@@ -62,7 +62,7 @@ func (gui *Gui) layout(g *gocui.Gui) error {
 	g.Highlight = true
 	width, height := g.Size()
 
-	information := "lazydocker " + gui.Config.Version
```
jesseduffield, line 65: no need to tell users that they're using lazydocker

pkg/gui/gui.go
```diff
@@ -360,7 +360,7 @@ func (gui *Gui) handleDonate(g *gocui.Gui, v *gocui.View) error {
 	if cx > len(gui.Tr.Donate) {
 		return nil
 	}
-	return gui.OSCommand.OpenLink("https://donorbox.org/lazydocker")
+	return gui.OSCommand.OpenLink("https://github.com/sponsors/jesseduffield")
```
jesseduffield, line 363: didn't realise I was still linking to donorbox

PR #343, opened 2022-05-12
exec attach via SDK

jesseduffield:
This is adapted from @skanehira 's docui repo which is no longer maintained (thanks for the help!)

There are two issues preventing merge:
1) I'm not sure if windows users will have issues given that I see some commented out code for handling resizing on windows
2) upon returning to lazydocker from the subprocess, a keypress is lost i.e. the next key you press is ignored. I've spent a while trying to get to the bottom of that with no luck. Interestingly the same issue occurs in the docui repo.


jesseduffield:
closing cos I was not able to get this to work

Issue #263, opened 2021-07-21
Resize shortcut


strowk:
**Is your feature request related to a problem? Please describe.**
Sometimes I need to see list of containers with their images, but size of screen does not allow to see both panes wide enough. 

**Describe the solution you'd like**
In that case I would like ability to resize panes with shortcuts, something like in tmux (I mean oh-my-tmux additional bindings).

**Describe alternatives you've considered**
Hiding right pane (with container details) could also work, but have a bit less application. Resizing could be nice just for better view in different situations, hiding is like resizing, but more ultimate and also probably would have to take more corner cases into consideration, like what scrolling shortcut would have to do when the right pane is hidden. 

**Additional context**
picture is showing how I have container with long name and cannot see its image (or other container as well) because there is just not enough place, but I do not ATM need to look at the logs and could resize/hide them

![image](https://user-images.githubusercontent.com/25692644/126490703-f1f56a7d-37bf-4fb1-b09d-d0ddb4c130b6.png)

jesseduffield:
This is done in https://github.com/jesseduffield/lazydocker/pull/372

jesseduffield:
https://github.com/jesseduffield/lazydocker/pull/372 is now in master. Lemme know if you want to re-open.

PR #352, opened 2022-06-03
Include image name on config tab

rmtsrc:
Sometimes the container image names can be cut off by long container names, for example in https://github.com/jesseduffield/lazydocker/issues/263.

This PR add the container name on the config tab:

Screenshot from 2022-06-03 17-13-14


jesseduffield:
Thanks @rmtsrc !

Issue #270, opened 2021-09-19
Use a home-based path for the Linux installation script


Eonfge:
**Is your feature request related to a problem? Please describe.**
The current Linux-installation script requires sudo-rights, and it places the file into the `/usr/local/bin`. This is not really necessary, since a user could also just install the script in `~/.local/bin` without sudo-rights.

Since users don't have any managed way to update this script, I would recommend against placing it in the system folder, as it might cause issues in the future. Besides, piping curl and bash is a bad practice and forcing users to do so with sudo-rights is even more dangerous.

**Describe the solution you'd like**
I would change the installation script and install to `~/.local/bin` by default

**Describe alternatives you've considered**
You could make a script-based switch: Sudo-rights: yes, install in `/usr/local/bin`. no, install in `~/.local/bin`. I would opt against this though, since it would still encourage sudo-curl-bash piping.

**Additional context**

Security discussion pages:
- https://blog.dijit.sh//don-t-pipe-curl-to-bash
- https://www.arp242.net/curl-to-sh.html
- and many more, but let's not digress

bergerom:
It seems that the Readme.md has not been updated with the new installation path (it is still written `/usr/local/bin`)

TheAxelander:
In it's current version the install script doesn't work if you run it directly with root (that's at least my experience). Yes, it installs lazydocker to `~/.local/bin` (which would be the same like `/root/.local/bin`) but then you are still unable to execute it globally. At the end I had to move it manually to `/usr/local/bin`.

In addition you always have a "left-over" lazydocker in the folder from where you have executed the script. I think that's coming from the change of `sudo mv` to `install` (line 23)

jesseduffield:
I'm happy to accept a PR to address this issue

pippo73:
Hello, 
I've discovered today your app, but I would like to notify that if you don't have "~/.local/bin" in your PATH variable, you are not able to use it ;-)
Maybe you should put a test in the installation script and ask the user where to put it or at least add an echo string to notify you should add this directory to the PATH ;-)

pippo73:
Hello, 
I've discovered today your app, but I would like to notify that if you don't have "~/.local/bin" in your PATH variable, you are not able to use it ;-)
Maybe you should put a test in the installation script and ask the user where to put it or at least add an echo string to notify you should add this directory to the PATH ;-)

PR #355, opened 2022-06-14
remove extracted lazydocker binary

DasFaultier:
The installation method used up until now would always leave one copy of the lazydocker binary in the extraction directory. This copy is not actually needed after installation, so it should be removed. TheAxelander also pointed this out in his comment at https://github.com/jesseduffield/lazydocker/issues/270#issuecomment-1042983635.
This commit changes that behaviour and makes the install script cleanup after itself.


jesseduffield:
LGTM!

PR #357, opened 2022-06-24
Add french doc

clemsau:
Add french translation for the keybindings doc.

In the `CONTRIBUTING.md` it is written that one should first discuss with the author before contributing, but considering how basic adding a translation to docs is, I assumed I could directly go for a PR.

Please decline the PR if the changes are unwanted.

Regards.


jesseduffield:
@clemsau keybindings are generated via `go run scripts/generate_cheatsheet.go`, and depend on a file in the `pkg/i18n/` directory for french. So I would create a french.go file in that directory.

jesseduffield (changes requested):
Marking as changes requested for my own bookkeeping

jesseduffield:
@clemsau sweet :) Looks like you'll need to rebase onto master as there were some changes since you first raised this PR

PR #358, opened 2022-06-28
Add Compose V2 instructions to FAQ

Syriiin:
Just ran into this when setting up a new environment with Compose V2.
As Compose V1 is now deprecated, perhaps lazydocker should default to the new `docker compose` command?
Could be too soon to make this change though, as many people are still using older versions of docker that don't have Compose V2 support.


jesseduffield:
We ought to just change the default to `docker compose` given most people should have that by now, and then have a doc explaining how to change it back to `docker-compose`. I'll create an issue for that.

jesseduffield:
Nonetheless let's merge this!

Issue #214, opened 2020-04-14
White font on white background in dialogs


DannyBen:
I suggest avoiding the use of white color for dialog fonts. It is barely readable for light background users.

Screenshot:

![image](https://user-images.githubusercontent.com/2405099/79246385-e7058100-7e81-11ea-8b8d-9753dc765a0b.png)

In addition, as you can see - I am not sure why the border, when active, changes its character set.

- I looked at the [config documentation](https://github.com/jesseduffield/lazydocker/blob/master/docs/Config.md) and couldn't see anything that helps for this particular issue.
- I saw there is a closed related issue #9 
- I saw there is an open request for more color customization (#67) - and although I would also love to see it, I believe the default configuration should be plain and easy (can be achieved by just not using black and white colors).

Lastly - another small thing I wish would be different in the UI, and does not deserve its own ticket - is the fact that selected rows are just bold, and not, say green or some other color. The bold is not that different than the non-bold in some terminal fonts.

eugenevd:
I'm using urxvt with solarized light theme and get something very similar to @dusty-phillips 
In lazygit I sorted this with lightTheme, would be nice here

jesseduffield:
Fixed by https://github.com/jesseduffield/lazydocker/pull/362

PR #362, opened 2022-08-16
Changed "White" colors to "Default" to work in both light and dark terminals

tilliwilli:
I just tried to quickly fix the issue that, in a light terminal, the text in menus is barely legible. See also https://github.com/jesseduffield/lazydocker/issues/214 on that.

I'm using Solarized Light/Dark and automatically switch between them via MacOS Dark Mode, and now with the font color set to "Default" instead of "White", it's easily readable in both modes. Also all the other themes that come with iTerm2 are looking good now.


jesseduffield:
Nice one, thanks for making this :)

PR #365, opened 2022-09-02
Fix execute shell command to take into account edge cases

### LINKED ISSUE #350: Shell is not executed for some containers
**Describe the bug**
Shell is not executed for some containers

**To Reproduce**
1. Pull `bitnami/openldap` image with `docker pull bitnami/openldap`
2. Launch it (`docker run --rm bitnami/openldap`)
3. Launch lazydocker
4. Select running openldap container and press `E` to get into container shell

**Expected behaviour**
You are dropped into container shell.

**Actual behaviour**
I see `id: cannot find name for user ID 1001` message.

This happens because image maintainer used `USER 1001` command [in the Dockerfile](https://github.com/bitnami/bitnami-docker-openldap/blob/master/2.6/debian-10/Dockerfile#L26) to set custom uid for the image. Obviously this user id is not present in `/etc/passwd` file - hence lazydocker command fails to detect shell.

I think lazydocker should try to detect preferred shell and if it fails to find one or execute one `/bin/sh` should be launched instead.

This approach will solve another shell-related problem (steps to reproduce below)
1. Pull `quay.io/keycloak/keycloak:18.0` image with `docker pull quay.io/keycloak/keycloak:18.0`
2. Launch it (`docker run --rm quay.io/keycloak/keycloak:18.0 start-dev`)
3. Launch lazydocker
4. Try to drop into shell by pressing `E` against keycloak container

You will see an error because Red Hat developers put `/sbin/nologin` (`/bin/false` seems to be another popular option in such cases though) as the shell for image user to tighten security. However, you still can directly exec `/bin/sh` and get access to the shell.

Also I should mention that I have an image without grep in it which also makes `exec shell` command to fail. Yes, it is a very specific image and I have custom command which simply execs `/bin/sh` against container but I would really like to see some bulletproof solution implemented in lazydocker.

jtraub:
This PR resolves #350 

The new proposed command is

```
/bin/sh -e -c 'eval $( (grep ^[^:]*:[^:]*:$(id -u): /etc/passwd || echo "::::::/bin/sh") | cut -d : -f 7-) || eval $(echo "/bin/sh")'
```

Changes are following:
1. Grep `/etc/passwd` with UID instead of username. `id -un` can start throwing errors if current user does not exist in `/etc/passwd` because it looks into the passwd file to determine user name (switch `-n`). So the first step is to make sure `id` doesn't throw - hence I changed grepping pattern to rely on uid (which comes in the [third field of the file](https://en.wikipedia.org/wiki/Passwd#Password_file)). While this change is not strictly required but it prevents output from error pollution. 

2. `|| echo "::::::/bin/sh` makes sure if grep doesn't find uid in passwd file we will feed a dummy entry to the pipe so lazydocker would start `/bin/sh` at least

3. The last part (`|| eval $(echo "/bin/sh")`) is here to guard against `/bin/false` or `/sbin/nologin` shells that are commonly used to tighten security. If we fail to launch shell specified in the `/etc/passwd` then lazydocker would start `/bin/sh` instead.

1 and 2 address problem with `bitnami/openldap` image mentioned in #350
In this case everything works flawlessly

3 addresses problem with `quay.io/keycloak/keycloak:18.0` image described in the same issue. Here execute shell works but error message that `/sbin/nologin` is not found yet present. I don't think it worth it to suppress the error as the command will become even more complicated

image

### DIFF
--- a/pkg/gui/containers_panel.go
+++ b/pkg/gui/containers_panel.go
@@ -533,7 +533,7 @@ func (gui *Gui) containerExecShell(container *commands.Container) error {
 	})
 
 	// TODO: use SDK
-	resolvedCommand := utils.ApplyTemplate("docker exec -it {{ .Container.ID }} /bin/sh -c 'eval $(grep ^$(id -un): /etc/passwd | cut -d : -f 7-)'", commandObject)
+	resolvedCommand := utils.ApplyTemplate("docker exec -it {{ .Container.ID }} /bin/sh -c 'eval $( (grep ^[^:]*:[^:]*:$(id -u): /etc/passwd || echo \"::::::/bin/sh\") | cut -d : -f 7-) || eval $(echo \"/bin/sh\")'", commandObject)
 	// attach and return the subprocess error
 	cmd := gui.OSCommand.ExecutableFromString(resolvedCommand)
 	return gui.runSubprocess(cmd)


jtraub:
I have no idea why lint fails. It fails to process `pkg/gui/confirmation_panel.go` file which I haven't touched.

mark2185:
> I have no idea why lint fails. It fails to process `pkg/gui/confirmation_panel.go` file which I haven't touched.

You can disregard that, it says "Unchanged files with check annotations", so it's not really relevant for this PR.

jesseduffield:
@jtraub if you rebase onto master those linting issues are fixed.

As for this PR, I'm a little hesitant about increasing the complexity of the existing command more for a couple bespoke images. Are these images representative of a larger set of images that have the same problems, or are they once-offs?

jtraub:
> Are these images representative of a larger set of images that have the same problems, or are they once-offs?

@jesseduffield I guess this depends on the area you are dealing with. For example, I encountered it a lot when I worked with LDAP-related images during spring (this is when I created the issue). I encountered images with `USER` command again in August when I started working on some datascience problems - this time it was unbearable so I decided to fix the problem for myself :-)

> if you rebase onto master those linting issues are fixed.

Is there a point in doing so? If you don't want to accept the changes I would prefer to keep it the code as is since my current lazydocker binary has been built from it and it works just fine for my needs.

PR #367, opened 2022-09-20
Add tab and backtab navigation

blennster:
mentioned in [#288](https://github.com/jesseduffield/lazydocker/pull/288#issuecomment-1008244271).

We could introduce a binding description but I think it would be better to not since it could become cluttery. As of now it will only change the main panes which works best for my workflow but let me know if you disagree.


jesseduffield:
Thanks @blennster !

PR #372, opened 2022-10-10
Various improvements

jesseduffield:
image

This PR does a few things:
* displays ports against containers/services
* allows cycling through window sizes (like in lazygit) via '+' and '_'
* fixes issues relating to focus
* shows selected line in blue

Given these ports can be long, we needed a way to increase the size of the current window so that we can see the whole port if necessary. I've split out a lazycore [repo](https://github.com/jesseduffield/lazycore) for shared functionality between lazygit and lazydocker and I've moved the boxlayout package into that.

I've also fixed up some focus issues I noticed along the way.

If the IP of the port is 0.0.0.0 we'll omit that, otherwise we'll show it


pkg/gui/confirmation_panel.go
```diff
@@ -15,25 +15,27 @@ import (
 
 func (gui *Gui) wrappedConfirmationFunction(function func(*gocui.Gui, *gocui.View) error) func(*gocui.Gui, *gocui.View) error {
 	return func(g *gocui.Gui, v *gocui.View) error {
+		if err := gui.closeConfirmationPrompt(); err != nil {
```
jesseduffield, line 18: following lazygit's lead by always returning focus BEFORE running a handler, to prevent all kinds of shenanigans

PR #373, opened 2022-10-10
Add message that cheatsheets are auto-generated

### LINKED ISSUE #369: Make it clear that cheetsheets are auto-generated
Sometimes somebody will raise a PR making manual changes to a cheatsheet file. These cheatsheet files are actually auto-generated and depend on files in pkg/i18n.

In [lazygit](https://github.com/jesseduffield/lazygit/blob/master/pkg/cheatsheet/generate.go#L71) we print a message at the top of the cheatsheet saying that it's auto-generated i.e.:
```go
		content = fmt.Sprintf("_This file is auto-generated. To update, make the changes in the "+
			"pkg/i18n directory and then run `%s` from the project root._\n\n%s", CommandToRun(), content)
```

We should do that here too in https://github.com/jesseduffield/lazydocker/blob/master/scripts/generate_cheatsheet.go

gusandrioli:
This PR resolves #369 

## Changes

Add message on top of each language's file specifying that cheatsheets are generated.

### DIFF
--- a/scripts/generate_cheatsheet.go
+++ b/scripts/generate_cheatsheet.go
@@ -40,6 +40,12 @@ func main() {
 
 		bindingSections := getBindingSections(mApp)
 		content := formatSections(mApp, bindingSections)
+		content = fmt.Sprintf(
+			"_This file is auto-generated. To update, make the changes in the "+
+				"pkg/i18n directory and then run `%s` from the project root._\n\n%s",
+			commandToRun(),
+			content,
+		)
 		writeString(file, content)
 	}
 }
@@ -133,4 +139,8 @@ func formatSections(mApp *app.App, bindingSections []*bindingSection) string {
 	}
 
 	return content
+}
+
+func commandToRun() string {
+	return "go run scripts/generate_cheatsheet.go"
 }
--- a/scripts/generate_cheatsheet.go
+++ b/scripts/generate_cheatsheet.go
@@ -19,6 +19,10 @@ import (
 	"github.com/jesseduffield/lazydocker/pkg/i18n"
 )
 
+const (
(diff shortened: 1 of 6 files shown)


scripts/generate_cheatsheet.go
```diff
@@ -40,6 +40,12 @@ func main() {
 
 		bindingSections := getBindingSections(mApp)
 		content := formatSections(mApp, bindingSections)
+		content = fmt.Sprintf(
+			"_This file is auto-generated. To update, make the changes in the "+
+				"pkg/i18n directory and then run `%s` from the project root._\n\n%s",
+			commandToRun(),
+			content,
+		)
```
gusandrioli, line 48: I followed `lazygit`'s pattern, but probably a slightly cleaner solution would be to convert `go run scripts/generate_cheatsheet.go` into a `const`. Let me know if you agree and i can update the PR.

gusandrioli:
@jesseduffield I noticed some extra changes on the keybindings files, let me know if it's something version-wise that i'm missing.

scripts/generate_cheatsheet.go
jesseduffield: @gusandrioli I'm happy for that to be converted to a const

jesseduffield:
@gusandrioli those extra changes are likely intentional, so we're all good there

scripts/generate_cheatsheet.go
gusandrioli: Updated!

gusandrioli:
@jesseduffield Awesome, thanks. PR updated.

jesseduffield:
great work @gusandrioli !

PR #378, opened 2022-10-11
various changes

jesseduffield:
supports:
* pressing home key to jump to top of main panel contents
* upping, downing project with 'U' and 'D' from the services panel

fixes:
* port not showing correctly when port not exposed on host machine


pkg/commands/container.go
```diff
@@ -57,6 +57,10 @@ func (c *Container) GetDisplayStrings(isFocused bool) []string {
 
 func (c *Container) displayPorts() string {
 	portStrings := lo.Map(c.Container.Ports, func(port types.Port, _ int) string {
+		if port.PublicPort == 0 {
```
jesseduffield, line 60: if port is not exposed on host machine we just want to show the container's port

Issue #354, opened 2022-06-13
Custom command passing a string literal


pythoninthegrass:
**Describe the bug**
Running an ad hoc `docker run` command in the terminal works. Passing the same command to lazydocker as either a scalar (folded/literal) or inline fails.

**To Reproduce**
1. Create a custom command in `config.yml`
    ```yaml
    - name: run-meetup
      attach: true
      command: >
        docker run -it --name meetup_bot
        -p 80:3000
        -e PORT="3000"
        -v "$(pwd)":/home/appuser/app
        --env-file .env
        meetup_bot
        bash
    ```
    or
    ```yaml
    - name: run-meetup
      attach: true
      command: 'docker run -it --name meetup_bot -p 80:3000 -e PORT="3000" -v "$(pwd)":/home/appuser/app --env-file .env meetup_bot bash'
    ```
2. Open lazydocker
3. Run the command
4. Raises error:
    ```bash
    + docker run -it --name meetup_bot -p 80:3000 -e PORT=3000 -v $(pwd):/home/appuser/app --env-file .env meetup_bot bash

    docker: Error response from daemon: create $(pwd): "$(pwd)" includes invalid characters for a local volume name, only "[a-zA-Z0-9][a-zA-Z0-9_.-]" are allowed. If you intended to pass a host directory, use absolute path.
    See 'docker run --help'.
    ```

**Expected behaviour**
It should interpolate/expand the `pwd` command or `$PWD` built-in variable in the command. Only able to hard-code the absolute path to work around the issue. Less than ideal, of course.

**Screenshots**
N/A

**Desktop (please complete the following information):**
 - OS: macOS 12.4 (21F79)
 - Lazydocker Version: v0.18.1
 - The last commit id if you built project from sources: N/A

**Additional context**
Using the new [docker mount syntax](https://docs.docker.com/storage/bind-mounts/) got the same results.

skanehira:
`$(pwd)` have to run in the shell, but lazydocker seems like run command directly.
I think this is not bug, but run commands through shell will be more convenient I think.

pythoninthegrass:
What would you suggest the workflow should be if this isn't a bug @skanehira?

jesseduffield:
I agree, we ought to run all custom commands in a shell so that things like $(pwd) can be evaluated. We do that currently with lazygit

jesseduffield:
[...]

PR #387, opened 2022-10-15
Wrap custom commands on shell

gusandrioli:
## Context
This solves https://github.com/jesseduffield/lazydocker/issues/354

## Implementation
Ideally, we'd re-use `ICmdObjBuilder` from `lazygit` across both projects, but that'd require a complete refactor here. So i ported the string replacement for Windows OS and the wrapping of new shell.

### DIFF
--- a/pkg/commands/os.go
+++ b/pkg/commands/os.go
@@ -22,13 +22,11 @@ import (
 
 // Platform stores the os state
 type Platform struct {
-	os                   string
-	shell                string
-	shellArg             string
-	escapedQuote         string
-	openCommand          string
-	openLinkCommand      string
-	fallbackEscapedQuote string
+	os              string
+	shell           string
+	shellArg        string
+	openCommand     string
+	openLinkCommand string
 }
 
 // OSCommand holds all the os commands
@@ -98,6 +96,25 @@ func (c *OSCommand) ExecutableFromStringContext(ctx context.Context, commandStr 
 	return exec.CommandContext(ctx, splitCmd[0], splitCmd[1:]...)
 }
 
+func (c *OSCommand) NewCommandStringWithShell(commandStr string) string {
+	quotedCommand := ""
+	// Windows does not seem to like quotes around the command
+	if c.Platform.os == "windows" {
+		quotedCommand = strings.NewReplacer(
+			"^", "^^",
+			"&", "^&",
+			"|", "^|",
+			"<", "^<",
+			">", "^>",
+			"%", "^%",
+		).Replace(commandStr)
+	} else {
+		quotedCommand = c.Quote(commandStr)
+	}
+
+	return fmt.Sprintf("%s %s %s", c.Platform.shell, c.Platform.shellArg, quotedCommand)
+}
+
 // RunCommand runs a command and just returns the error
 func (c *OSCommand) RunCommand(command string) error {
 	_, err := c.RunCommandWithOutput(command)
@@ -114,16 +131,6 @@ func (c *OSCommand) FileType(path string) string {
 		return "directory"
 	}
 	return "file"
-}
-
-// RunDirectCommand wrapper around direct commands
-func (c *OSCommand) RunDirectCommand(command string) (string, error) {
-	c.Log.WithField("command", command).Info("RunDirectCommand")
-
-	return sanitisedCommandOutput(
-		c.command(c.Platform.shell, c.Platform.shellArg, command).
-			CombinedOutput(),
-	)
 }
 
 func sanitisedCommandOutput(output []byte, err error) (string, error) {
@@ -190,12 +197,23 @@ func (c *OSCommand) PrepareSubProcess(cmdName string, commandArgs ...string) *ex
 
 // Quote wraps a message in platform-specific quotation marks
 func (c *OSCommand) Quote(message string) string {
-	message = strings.Replace(message, "`", "\\`", -1)
-	escapedQuote := c.Platform.escapedQuote
-	if strings.Contains(message, c.Platform.escapedQuote) {
-		escapedQuote = c.Platform.fallbackEscapedQuote
+	var quote string
+	if c.Platform.os == "windows" {
+		quote = `\"`
+		message = strings.NewReplacer(
+			`"`, `"'"'"`,
+			`\"`, `\\"`,
+		).Replace(message)
+	} else {
+		quote = `"`
+		message = strings.NewReplacer(
+			`\`, `\\`,
+			`"`, `\"`,
+			`$`, `\$`,
+			"`", "\\`",
+		).Replace(message)
 	}
-	return escapedQuote + message + escapedQuote
+	return quote + message + quote
 }
 
 // Unquote removes wrapping quotations marks if they are present
--- a/pkg/commands/os_test.go
+++ b/pkg/commands/os_test.go
@@ -1,6 +1,7 @@
 package commands
 
 import (
+	"fmt"
 	"io/ioutil"
 	"os"
 	"os/exec"
@@ -153,13 +154,14 @@ func TestOSCommandEditFile(t *testing.T) {
 	}
 }
 
-// TestOSCommandQuote is a function.
 func TestOSCommandQuote(t *testing.T) {
 	osCommand := NewDummyOSCommand()
 
+	osCommand.Platform.os = "linux"
+
 	actual := osCommand.Quote("hello `test`")
 
-	expected := osCommand.Platform.escapedQuote + "hello \\`test\\`" + osCommand.Platform.escapedQuote
+	expected := "\"hello \\`test\\`\""
 
 	assert.EqualValues(t, expected, actual)
 }
@@ -169,12 +171,10 @@ func TestOSCommandQuoteSingleQuote(t *testing.T) {
 	osCommand := NewDummyOSCommand()
 
 	osCommand.Platform.os = "linux"
-	osCommand.Platform.fallbackEscapedQuote = "\""
-	osCommand.Platform.escapedQuote = "'"
 
 	actual := osCommand.Quote("hello 'test'")
 
-	expected := osCommand.Platform.fallbackEscapedQuote + "hello 'test'" + osCommand.Platform.fallbackEscapedQuote
+	expected := `"hello 'test'"`
 
 	assert.EqualValues(t, expected, actual)
 }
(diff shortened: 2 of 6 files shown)


gusandrioli:
Here's the result using the issue's example: 
```sh
(base) ➜  lazydocker git:(master) ✗ go run -mod=vendor main.go
docker-compose config --quiet

+ bash -c docker run -it -v "$(pwd)":/data redis

1:C 15 Oct 2022 19:27:44.265 # oO0OoO0OoO0Oo Redis is starting oO0OoO0OoO0Oo
1:C 15 Oct 2022 19:27:44.265 # Redis version=7.0.5, bits=64, commit=00000000, modified=0, pid=1, just started
1:C 15 Oct 2022 19:27:44.265 # Warning: no config file specified, using the default config. In order to specify a config file use redis-server /path/to/redis.conf
1:M 15 Oct 2022 19:27:44.265 * monotonic clock: POSIX clock_gettime
                _._                                                  
           _.-``__ ''-._                                             
      _.-``    `.  `_.  ''-._           Redis 7.0.5 (00000000/0) 64 bit
  .-`` .-```.  ```\/    _.,_ ''-._                                  
 (    '      ,       .-`  | `,    )     Running in standalone mode
 |`-._`-...-` __...-.``-._|'` _.-'|     Port: 6379
 |    `-._   `._    /     _.-'    |     PID: 1
  `-._    `-._  `-./  _.-'    _.-'                                   
 |`-._`-._    `-.__.-'    _.-'_.-'|                                  
 |    `-._`-._        _.-'_.-'    |           https://redis.io       
  `-._    `-._`-.__.-'_.-'    _.-'                                   
 |`-._`-._    `-.__.-'    _.-'_.-'|                                  
 |    `-._`-._        _.-'_.-'    |                                  
  `-._    `-._`-.__.-'_.-'    _.-'                                   
      `-._    `-.__.-'    _.-'                                       
          `-._        _.-'                                           
              `-.__.-'                                               

1:M 15 Oct 2022 19:27:44.266 # Server initialized
1:M 15 Oct 2022 19:27:44.269 * Ready to accept connections
^C1:signal-handler (1665862067) Received SIGINT scheduling shutdown...
1:M 15 Oct 2022 19:27:47.999 # User requested shutdown...
1:M 15 Oct 2022 19:27:47.999 * Saving the final RDB snapshot before exiting.
1:M 15 Oct 2022 19:27:48.022 * DB saved on disk
1:M 15 Oct 2022 19:27:48.023 # Redis is now ready to exit, bye bye...


Press enter to return to lazydocker (this prompt can be disabled in your config by setting `gui.returnImmediately: true`)
```

pkg/commands/os.go
```diff
@@ -83,6 +83,27 @@ func (c *OSCommand) ExecutableFromString(commandStr string) *exec.Cmd {
...
+			"^", "^^",
+			"&", "^&",
+			"|", "^|",
+			"<", "^<",
+			">", "^>",
+			"%", "^%",
+		).Replace(commandStr)
+	} else {
+		quotedCommand = c.Quote(commandStr)
+	}
```
gusandrioli, line 100: I copied this over from `lazygit`. Was there any reason this was not here before?

pkg/commands/os_test.go
```diff
@@ -291,3 +291,25 @@ func TestOSCommandCreateTempFile(t *testing.T) {
...
+	osCommand := NewDummyOSCommand()
+	tests := []struct {
+		name       string
+		commandStr string
+		want       []string
+	}{
+		{
+			"success",
+			"pwd",
+			[]string{osCommand.Platform.shell, osCommand.Platform.shellArg, "pwd"},
```
gusandrioli, line 305: Validates that it contains `bash -c ...`

gusandrioli:
@jesseduffield This is ready, let me know your thoughts!

pkg/commands/os.go
gusandrioli: Also, i didn't wrap the shell directly on `ExecutableFromString` because it was touching a lot of different functions and caused some issues. E.g. running `EditFile(filename string) (*exec.Cmd, error)` with it.

jesseduffield: No good reason. Also be sure to update the Quote method based on lazygit i.e.:
```go
func (self *CmdObjBuilder) Quote(message string) string {
	var quote string
	if self.platform.OS == "windows" {
		quote = `\"`
		message = strings.NewReplacer(
			`"`, `"'"'"`,
			`\"`, `\\"`,
		).Replace(message)
	} else {
		quote = `"`
		message = strings.NewReplacer(
			`\`, `\\`,
			`"`, `\"`,
			`$`, `\$`,
			"`", "\\`",
		).Replace(message)
	}
	return quote + message + quote
}
```

pkg/commands/os_test.go
jesseduffield: nice

jesseduffield (changes requested):
Looking good, one comment.

Also we should remove `RunDirectCommand` given that's not being used and seemed to do something similar

jesseduffield:
Also one thing we'll want to test @gusandrioli is whether an existing custom command which already used bash -c explicitly now breaks with this new code

gusandrioli:
> Also one thing we'll want to test @gusandrioli is whether an existing custom command which already used bash -c explicitly now breaks with this new code

Good point, i'll test that out.

gusandrioli:
@jesseduffield It does fail! Thanks for catching that: 
```sh
+ bash -c bash -c docker run -it -v "$(pwd)":/data redis


Usage:  docker [OPTIONS] COMMAND

A self-sufficient runtime for containers
```

One alternative is to leave the command as is if it contains `-c` . What are your thoughts on this?

pkg/commands/os.go
```diff
@@ -78,11 +76,40 @@ func (c *OSCommand) RunExecutable(cmd *exec.Cmd) error {
...
+
+	return c.ExecutableFromString(
+		fmt.Sprintf("%s %s %s", c.Platform.shell, c.Platform.shellArg, quotedCommand),
+	)
+}
+
+func cmdHasShellInvokation(cmd string) bool {
+	return strings.Contains(cmd, "bash -c") || strings.Contains(cmd, "sh -c")
+}
```
gusandrioli, line 111: Not sure if this covers all cases, but covers the most common shells. Covers:
- /bin/bash
- /bin/sh
- bash -c
- sh -c

jesseduffield: I'm not sure checking the string itself is a good approach: there's going to be edge cases we miss (e.g. custom shells like fish, zsh). I think we need to put this on the user somehow. We can add a `shell` key (boolean value) to the custom command config, which defaults to false. Unfortunately this means the user will need to actually set that to if they want to have it invoked from their shell, but that's the only way to ensure backwards compatibility right now.

Although I think 'true' would be a better default, I'm happy to make this change and then see if people complain and think it's worth breaking backwards compatibility to change the default

gusandrioli: Awesome, makes sense. I'll make that change.

pkg/config/app_config.go
```diff
@@ -306,6 +306,10 @@ type CustomCommand struct {
 	// option where the output plays in the main panel.
 	Attach bool `yaml:"attach"`
 
+	// InvokeShell indicates whether to invoke the Command on a shell or not.
+	// Example of a bash invoked command: `/bin/bash -c "{Command}".
+	InvokeShell bool `yaml:"invokeShell"`
```
gusandrioli, line 311: Let me know if the description looks fine.

pkg/gui/custom_commands.go
```diff
@@ -52,8 +54,11 @@ func (gui *Gui) createCommandMenu(customCommands []config.CustomCommand, command
 
 		// if we have a command for attaching, we attach and return the subprocess error
 		if option.customCommand.Attach {
-			cmd := gui.OSCommand.ExecutableFromString(option.command)
-			return gui.runSubprocess(cmd)
+			if option.invokeShell {
+				return gui.runSubprocess(gui.OSCommand.ExecutableFromStringWithShell(option.command))
+			}
+
+			return gui.runSubprocess(gui.OSCommand.ExecutableFromString(option.command))
```
gusandrioli, line 61: Also renamed the func to `ExecutableFromStringWithShell` to follow the pattern of `Executable` funcs on `pkg/commands/os.go`

gusandrioli (commented):
@jesseduffield Updated the config to have a `invokeShell` bool (default: `false`). Tested on both cases and worked fine. 

Also, for these type of PRs with a lot of commits, do you prefer if i squash them or leave as is?

pkg/gui/custom_commands.go
```diff
@@ -52,12 +52,12 @@ func (gui *Gui) createCommandMenu(customCommands []config.CustomCommand, command
...
+		}
+
 		// if we have a command for attaching, we attach and return the subprocess error
 		if option.customCommand.Attach {
-			if option.invokeShell {
-				return gui.runSubprocess(gui.OSCommand.ExecutableFromStringWithShell(option.command))
-			}
-
 			return gui.runSubprocess(gui.OSCommand.ExecutableFromString(option.command))
 		}
```
gusandrioli, line 62: Since we moved the fix for a user config level, i think it makes sense to allow users to run invoked shell cmds on attach true and false.

pkg/config/app_config.go
jesseduffield: I would be even more concise than that and just say `shell`

pkg/gui/custom_commands.go
```diff
@@ -50,10 +52,13 @@ func (gui *Gui) createCommandMenu(customCommands []config.CustomCommand, command
 			return option.customCommand.InternalFunction()
 		}
 
+		if option.invokeShell {
+			option.command = gui.OSCommand.NewCommandStringWithShell(option.command)
```
jesseduffield, line 56: rather than mutate option.command here I would have a separate `resolvedCommand` variable that we use. This means that if we want to reuse `options` from outside this callback down the line there's no chance of the callback's code interfering

jesseduffield (changes requested):
couple things

pkg/gui/custom_commands.go
gusandrioli: I see. Okay, makes sense.

jesseduffield:
Nice work :) I'll give this a local test soon

gusandrioli:
Hey @jesseduffield! Hope you're doing well. Are we still going to use this PR? If so, i can solve the conflicts.

jesseduffield:
@gusandrioli sorry this fell through the cracks. I'll take another look this afternoon

gusandrioli:
@jesseduffield No worries at all, let me know :)

jesseduffield:
Code looks good, I've just fixed some merge conflicts

jesseduffield:
can you confirm this still works on your end after I resolved those merge conflicts @gusandrioli ?

jesseduffield:
Merged. Thanks @gusandrioli !

mark2185:
This broke the custom commands like [these](https://github.com/jesseduffield/lazydocker/blob/master/docs/Config.md#custom-commands):
```
customCommands:
  containers:
    - name: bash
      attach: true
      command: 'docker exec -it {{ .Container.ID }} bash'
      serviceNames: []
```
Results with the error:
```
+ bash -c docker exec -it {{ .Container.ID }} /bin/bash

Error response from daemon: No such container: {{
```

Setting `shell: false` changes nothing.

VaultVulp:
This changeset broke `lazydocker` completely for me. I hope the issue will be resolved, ASAP.

For now I've downgrade to `0.20.0` via `homebrew` using the following snippet:
```bash
wget https://raw.githubusercontent.com/jesseduffield/homebrew-lazydocker/b6d29bcd59058d1b7ec371df12f75d7048b40789/lazydocker.rb
brew uninstall lazydocker
brew install -s lazydocker.rb
rm ./lazydocker.rb
```

Issue #385, opened 2022-10-13
Hotkey for entering a running container's shell


kbairak:
**Is your feature request related to a problem? Please describe.**
It is very useful to enter a running container's shell. A shortcut for this would be appreciated. I don't need this for myself as already have a workaround. I think it would be nice for others.

**Describe the solution you'd like**
A built-in hotkey that, when pressed on a service, enters the containers shell via `docker exec`.

**Describe alternatives you've considered**
Here is what I have in my config file:

```yaml
---
customCommands:
  containers:
    - name: shell
      attach: true
      command:
        docker exec -it {{ .Container.ID }} sh -c 'for SH in $(echo bash sh); do if (eval $SH); then exit 0; fi; done || exit 1'
      serviceNames: []
```

(My bash scripting skills is not something I boast about)

This way, pressing `c` on a running service gives me the "shell" option I can use to enter a container's shell.

**Additional context**
_If_ this were to be implemented and documented, a note that the service must be configured with `stdin_open` and `tty` in order for this to work would be helpful.

mark2185:
@kbairak Isn't that behind the `E` key? 

The menu states `E    exec shell`

kbairak:
I hadn't discovered this before. Still, it only works on the "Standalone Containers" window, no the "Services" window.

PR #389, opened 2022-10-19
Add 'E' keybinding (exec shell) to services panel

kbairak:
Addresses https://github.com/jesseduffield/lazydocker/issues/385

### DIFF
--- a/pkg/gui/keybindings.go
+++ b/pkg/gui/keybindings.go
@@ -269,6 +269,13 @@ func (gui *Gui) GetInitialKeybindings() []*Binding {
 			Description: gui.Tr.ExecShell,
 		},
 		{
+			ViewName:    "services",
+			Key:         'E',
+			Modifier:    gocui.ModNone,
+			Handler:     gui.handleContainersExecShell,
+			Description: gui.Tr.ExecShell,
+		},
+		{
 			ViewName:    "containers",
 			Key:         'c',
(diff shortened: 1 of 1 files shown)


pkg/gui/keybindings.go
```diff
@@ -268,6 +268,13 @@ func (gui *Gui) GetInitialKeybindings() []*Binding {
 			Handler:     gui.handleContainersExecShell,
 			Description: gui.Tr.ExecShell,
 		},
+		{
+			ViewName:    "services",
+			Key:         'E',
+			Modifier:    gocui.ModNone,
+			Handler:     gui.handleContainersExecShell,
```
jesseduffield, line 275: I believe this will shell into the selected container in the containers panel, but we want it to shell into the container of the selected service in the services panel. Could you create a new function for that in `services_panel.go`?

jesseduffield (changes requested):
one thing :)

pkg/gui/keybindings.go
kbairak: Actually, I just discovered that `E` works on the services panel too :confused: . I don't know how this happened, perhaps my package manager had an older version and then it was fixed. So I think this PR doesn't make any sense anymore.

kbairak:
@jesseduffield Thanks for taking the time to check this out and answer me.

PR #390, opened 2022-10-21
deterministically display container ports

jesseduffield:
Turns out that when a container has multiple ports, they're not deterministically ordered, which causes an annoying flickering effect as they're constantly swapped.

This fixes that by sorting alphabetically.

### DIFF
--- a/pkg/commands/container.go
+++ b/pkg/commands/container.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"os/exec"
+	"sort"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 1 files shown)


pkg/commands/container.go
```diff
@@ -58,7 +59,7 @@ func (c *Container) GetDisplayStrings(isFocused bool) []string {
 func (c *Container) displayPorts() string {
 	portStrings := lo.Map(c.Container.Ports, func(port types.Port, _ int) string {
 		if port.PublicPort == 0 {
-			return fmt.Sprintf("%d\\%s", port.PrivatePort, port.Type)
+			return fmt.Sprintf("%d/%s", port.PrivatePort, port.Type)
```
jesseduffield, line 62: accidentally had this as a backslash previously

PR #392, opened 2022-10-23
List panel filtering

jesseduffield:
This PR adds the ability to filter on side panels by pressing '/' and typing in some text.

There are many other things this PR does that I added in the process:
* refactoring to DRY up the code by introducing a SideListPanel struct
* adding deadlock-detecting mutexes when in debug mode
* fixed a bug in gocui related to horizontal scrolling
* added gui.Menu function for creating menus (following Lazygit's lead)

One of the purposes of my refactor is to make it easy to add another side panel, as I expect a networks panel will be added soon enough.


jesseduffield:
I'm gonna test this myself locally for a few days before cutting a release

PR #408, opened 2022-11-26
better screen mode approach

jesseduffield:


tjschuck:
Sorry to resurrect a very old PR, but this is the only mention in the repo I see of the `tall` screen mode, which seems like exactly what I want 😅

It doesn't seem to work on the current version, so I'm just wondering if it was a legacy screen mode that was removed?  Or if there's something else I'm missing.  (Most optimistically, I'd like to open `lazydocker` with `tall` mode and the `services` panel focused by default.)

PR #419, opened 2022-12-23
Short status

martini1992:
- Add ability to use a 'short' or 'icon' set for showing container status.
- Some additional refactoring of util functions to make what the variables are more clear and improve (not fix, go-runewidth isn't bulletproof) support for emoji and unicode text symbols in tables. (This can be rolled back if people aren't on-board).
- Added config to allow it to be set from 'long' (default), 'short' and 'icon'.
- Added to the Config.md to show it can be configured.


jesseduffield:
Looks good. I pushed some small changes

PR #424, opened 2023-01-16
Add "networks" panel under "volumes"

Tony-Sol:
Just a little improvement with adding "Networks" panel under "Volumes" one

half_mode
normal_mode


jesseduffield:
Thanks for making this! Sorry I've been super busy lately but I'll give this a review when I get the chance

pkg/gui/networks_panel.go
```diff
@@ -0,0 +1,185 @@
...
+		ListPanel: panels.ListPanel[*commands.Network]{
+			List: panels.NewFilteredList[*commands.Network](),
+			View: gui.Views.Networks,
+		},
+		NoItemsMessage: gui.Tr.NoNetworks,
+		Gui:            gui.intoInterface(),
+		// we're sorting these networks based on whether they have labels defined,
+		// because those are the ones you typically care about.
+		// Within that, we also sort them alphabetically
+		Sort: func(a *commands.Network, b *commands.Network) bool {
```
jesseduffield, line 43: I assume we do something similar with another panel, but for networks is it actually true that we care more about labels than names? I suspect not

jesseduffield (changes requested):
Sorry for taking so long to review!

This PR is great. All the code looks good: I just have one question

pkg/gui/networks_panel.go
Tony-Sol: Yes, it's exactly like volumes sorting, which fits to networks, i guess.

So, may be should we sort networks by driver at first and then by names?

jesseduffield: I'm thinking we just sort by name: I doubt the typical user has many networks configured anyway

Tony-Sol: Done :)

jesseduffield:
Tested locally and it works like a charm.

Looks like you need to regenerate cheatsheets though (using `go run scripts/cheatsheet/main.go generate`)

jesseduffield:
Thanks for making this @Tony-Sol !

jessejoe:
Should we be seeing this in the `brew tap`? I don't see it currently.

PR #428, opened 2023-01-27
Docs: explain go templates in custom commands

jchvz:
This PR edits the `Config.md` doc to explain the go template variables that are available to authors of custom commands.

With this info documented, it becomes much easier for users to write commands such as:
```yml
customCommands:
  containers:
    -
      name: dive
      attach: true
      command: dive {{ .Container.Container.ImageID }}
      serviceNames: []
```

This also takes care of a minor yaml issue in the doc. Before the change, Github highlighted this as invalid:
image

### DIFF
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -104,6 +104,8 @@ customCommands:
       serviceNames: []
 ```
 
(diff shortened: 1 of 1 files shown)


jesseduffield:
Thanks for making this @jchvz !

PR #444, opened 2023-04-05
Allow to render container stats and configs both in `json` and `yaml`

Tony-Sol:
Add optional config `".gui.detailsFormat"` into config file, which provides ability to change output format of container stats and config details.

To preserve current behaviour, use `json` as default.

Keys and their order are same between `json` and `yaml` formats.

Снимок экрана 2023-04-05 в 22 00 21


ViktorKram:
Beast.

Tony-Sol:
@jesseduffield would you please take a look at this?

jesseduffield:
Sorry for the late review,

I actually much prefer the yaml appearance to the JSON appearance. I'm happy to make it yaml without it being configurable. Could you do that @Tony-Sol ?

Tony-Sol:
@jesseduffield mm, maybe just make yaml by default, but keep this setting?
I prefer yaml over json as well (that's why i made this PR 😉), but definitely there are users who like json more.

jesseduffield:
For rendering in a UI though, I think yaml makes way more sense. k9s for example uses yaml to display pod specs and to my knowledge doesn't offer a way of instead showing it in JSON. I reckon we just switch to yaml and make it configurable if people complain. In general I want to keep configuration to a minimum where possible as it's a maintenance burden

Tony-Sol:
Also, there is a way to make yaml output colorized, but it requires to add goccy/go-yaml dependency with other updates in go.mod

Should we make this or common white text is just fine?

colorizedStats
colorizedConfig

jesseduffield:
Nice, let's add that dependency, it's great for readability

Tony-Sol:
Should i make colors fully customizable, set some pre-configured themes or just hardcode colors?

jesseduffield:
hardcode :)

Tony-Sol:
I see a point of future improvements - make themes or something, to make ui look more smooth and less like christmas tree 😄 
(I'll make it as soon as get free time)

Tony-Sol:
So, this is how it looks like now
docker-compose_config
about
stats
config

pkg/utils/utils.go
```diff
@@ -339,3 +384,23 @@ func IsValidHexValue(v string) bool {
...
+func MarshalIntoFormat(data interface{}, format string) ([]byte, error) {
+	dataJSON, err := json.MarshalIndent(data, "", "  ")
+	if err != nil {
+		return nil, err
+	}
+	switch format {
+	case "json":
+		return dataJSON, err
+	case "yaml":
+		// Use Unmarshal->Marshal hack to convert vendor-locked objects to YAML with same structure as JSON
```
jesseduffield, line 397: Could you explain a bit more about what's happening here? What do you mean by vendor locked?

jesseduffield (commented):
Could you add a unit test for this?

pkg/utils/utils.go
```diff
@@ -52,6 +58,45 @@ func ColoredString(str string, colorAttribute color.Attribute) string {
 	return ColoredStringDirect(str, colour)
 }
 
+// ColoredYamlString takes an YAML formatted string and returns a colored string
+// with colors hardcoded as:
+// keys: cyan
+// Booleans: magenta
+// Numbers: yellow
+// Strings: green
+func ColoredYamlString(str string) string {
```
jesseduffield, line 67: Can you add a unit test for this? (Accidentally just left this as a general comment cos I'm reviewing from phone :) )

jesseduffield (commented):
Looking great! Left a couple comments. We also need to scrap the JSON-related code

pkg/utils/utils.go
Tony-Sol: Sure, `container.Details` declared in `vendor/github.com/docker/docker/api/types/stats.go` already has json tags, but not yaml.

So, in order to keep `./vendor/` files consistent and not changed in any inappropriate way, like
```diff
-- CPUUsage CPUUsage `json:"cpu_usage"`
++ CPUUsage CPUUsage `json:"cpu_usage" yaml:"cpu_usage"`
```
and keep the resulting yaml structure exactly as json is.

pkg/utils/utils.go
Tony-Sol: @jesseduffield sure, done

Tony-Sol: And sadly, i don't get about scrapping "the JSON-related code" :c

Tony-Sol:
@jesseduffield is everything good?, are we ready for merge?:)

pkg/utils/utils.go
jesseduffield: What I mean is, let's remove `DetailsFormat` and always use yaml formatting.

pkg/utils/utils.go
jesseduffield: Ah, that makes sense. Could you capture that in a comment in the code?

jesseduffield:
@Tony-Sol I've left a couple more comments :)

jesseduffield:
> Think it's better for possible improvements in future, hope you agree 😉

Yep that's fine.

Merged! Great work on this @Tony-Sol !

PR #460, opened 2023-07-19
Add shortcut (Ctrl+L) to clear logs 

### LINKED ISSUE #403: Clear logs
It'd be really useful to be able to clear the logs view and only show new items as they come in.

This is possible when viewing the logs independently (i.e. via the `m` keybind) and using the clear command to clear the terminal, but I often find myself jumping between containers to check the logs, so being able to do this within the lazydocker context would be handy!

sunny-Ne5:
Fix #403

### DIFF
--- a/pkg/gui/keybindings.go
+++ b/pkg/gui/keybindings.go
@@ -450,6 +450,13 @@ func (gui *Gui) GetInitialKeybindings() []*Binding {
 			Handler:  gui.scrollRightMain,
 		},
 		{
+			ViewName:    "main",
+			Key:         gocui.KeyCtrlL,
+			Modifier:    gocui.ModNone,
+			Handler:     wrappedHandler(gui.handleClearMain),
+			Description: gui.Tr.ClearMain,
+		},
+		{
 			ViewName: "main",
 			Key:      'h',
 			Modifier: gocui.ModNone,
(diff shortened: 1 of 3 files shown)


jesseduffield:
Looking good. Two things:
* Could we make this keybinding also available in the side views that render to the main view? I'd prefer to not have to hit enter in order to press ctrl+l
* Looks like the keybinding isn't rendered properly in the keybindings menu (which you get upon pressing '?'
image

Looks like we need to add a special key to the switch statement in GetKey() in pkg/gui/keybindings.go

sunny-Ne5:
> * Could we make this keybinding also available in the side views that render to the main view? I'd prefer to not have to hit enter in order to press ctrl+l

Should we enable this keybinding for "Containers" or for other side views as well? For this I was planning to add one more entry to GetInitialKeybindings() in keybindings.go. Let me know if there is a better way to do it if it's repetitive.

>*Looks like the keybinding isn't rendered properly in the keybindings menu (which you get upon pressing '?'
Is this good?
image

jesseduffield:
@sunny-Ne5 Sorry for the late reply! You'll only need the keybinding for containers and services, so feel free to just add an extra two keybindings for that

peauc:
I want to take a look at this PR to move it forward.
@sunny-Ne5, are you still interested in this PR, you don't mind ?

sunny-Ne5:
please go ahead

peauc:
@jesseduffield requested changes have been made in my pull request 👍

obfox:
Do you added this feature on different branch?

peauc:
@obfox hello, sadly the repo is not maintained. All pull requests are merged on my personal fork if you want to get a build with a few fixes.

PR #473, opened 2023-08-11
i18n: spanish translation

luisnquin:
Currently the project has no **spanish** translation so here is some help for that

I think all the work I've done on that is good. I hope you enjoy this contribution!


jesseduffield:
Thanks for making this!

PR #476, opened 2023-08-28
Bump version tag in installation docs for asdf

helgeblod:
- Latest version is 0.21.1
- Update asdf install instructions to reflect latest version


jesseduffield:
Is it possible to use a 'latest' tag?

helgeblod:
Yes, that's even better. I'll update the PR 👍

helgeblod:
Updated the PR to use `latest` tag in asdf and confirmed that it works 👍

jesseduffield:
Thanks for making this

PR #501, opened 2023-11-22
feat: add volume `CreatedAt` field to volumes Config side panel

m-mead:
This MR adds the `CreatedAt` value associated with a Docker volume to the volumes Config side panel. Here is an example of how the value looks, and how it matches the output of `docker volume inspect <volume>`.

```
$ docker volume inspect vscode
[
    {
        "CreatedAt": "2022-11-07T21:06:17Z",
        "Driver": "local",
        "Labels": {},
        "Mountpoint": "/var/lib/docker/volumes/vscode/_data",
        "Name": "vscode",
        "Options": {},
        "Scope": "local"
    }
]
```

image

### DIFF
--- a/pkg/commands/volume.go
+++ b/pkg/commands/volume.go
@@ -17,6 +17,7 @@ type Volume struct {
 	OSCommand     *OSCommand
 	Log           *logrus.Entry
 	DockerCommand LimitedDockerCommand
+	CreatedAt     string
 }
 
 // RefreshVolumes gets the volumes and stores them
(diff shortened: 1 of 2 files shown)


m-mead:
After making the change, I did notice the Create field in the Images > Config side panel seem to be a formatted to the local timezone. I'm happy to make the change to do the same for the volume CreatedAt field for consistent look and feel. Or I can leave it as is. Either works for me.

Example of the Crated field in the Images > Config side panel
image

mark2185:
Given that you've closed the PR, are you against someone else taking this up and crediting you as the author?

m-mead:
> Given that you've closed the PR, are you against someone else taking this up and crediting you as the author?

Hey @mark2185, I closed it because I didn't open an issue first like the contributing docs ask, so I figured this would not be accepted. If this functionality is desired I'd prefer to see it through and can reopen the PR or create an issue for discussion.

PR #505, opened 2024-01-12
Fix install with unsupported tar -z

lars18th:
Some tar binaries (like the one from busybox in buildroot) have troubles with "tar -z" parameters. This patch overcomes the problem calling to gzip prior to tar.

### DIFF
--- a/scripts/install_update_linux.sh
+++ b/scripts/install_update_linux.sh
@@ -19,6 +19,7 @@ GITHUB_URL="https://github.com/jesseduffield/lazydocker/releases/download/${GITH
 
 # install/update the local binary
 curl -L -o lazydocker.tar.gz $GITHUB_URL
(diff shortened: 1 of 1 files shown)


mark2185:
Do they work if you don't specify `-z` at all? I know I've encountered some that worked with `tar xf`, even though it was gzipped.

lars18th:
> Do they work if you don't specify `-z` at all? I know I've encountered some that worked with `tar xf`, even though it was gzipped.

No. The "-z" parameter is not supported by busybox's "tar". Futhermore, some other UNIX tar versions doesn't support it. The alternative implemented using the "gzip" tool externally is the most universal way. So I suggest to merge it.

Regards.

lars18th:
Hi,

Added the missing "v" (verbose) parameter in the tar command.

I suggest to merge it.
Regards.

PR #534, opened 2024-05-26
Fix buggy log rendering

jesseduffield:
Two problems this PR solves:

## 1. Blank logs

Previously we would only load container details a second after we loaded the containers themselves, and we need those container details loaded to know whether a container uses a TTY or not. There are two separate methods you need to use when streaming container logs depending on whether it uses a TTY or not, and we were defaulting to assuming no TTY if the container details were not loaded. If this assumption was wrong, the streaming function would return an error and we would not retry so the logs view would just remain black until the user navigated away from the container and them came back.

Now we load the container details within the method that loads the containers themselves so we can expect them to be present. Interestingly enough, it seems that sometimes docker will return us empty details sometimes so to defend against that, I've also got a check on whether details are loaded, which loops until we get the details or the task is cancelled.

## 2. Lack of screen updates after returning from subprocess

After returning from a subprocess, the screen would only update occasionally: triggered by a docker event. That was because the background task which refreshes the screen every 30ms was killed entirely upon switching to the subprocess, and never recreated. Now we just pause it while the subprocess is still open.

### DIFF
--- a/pkg/gui/container_logs.go
+++ b/pkg/gui/container_logs.go
@@ -119,7 +119,24 @@ func (gui *Gui) writeContainerLogs(container *commands.Container, ctx context.Co
 	}
 	defer readCloser.Close()
 
-	if container.DetailsLoaded() && container.Details.Config.Tty {
+	if !container.DetailsLoaded() {
+		// loop until the details load or context is cancelled, using timer
+		ticker := time.NewTicker(time.Millisecond * 100)
+		defer ticker.Stop()
+	outer:
+		for {
+			select {
+			case <-ctx.Done():
+				return nil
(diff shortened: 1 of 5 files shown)


pkg/gui/container_logs.go
```diff
@@ -114,10 +114,29 @@ func (gui *Gui) writeContainerLogs(container *commands.Container, ctx context.Co
 		Follow:     true,
 	})
 	if err != nil {
+		gui.Log.Error(err)
 		return err
 	}
+	defer readCloser.Close()
```
jesseduffield, line 120: This is a separate thing: we should have been closing this from the start.

PR #542, opened 2024-06-06
Update Dockerfile

blark:
Update alpine version to 3.19 as 3.15 is no longer available, which breaks build. Tested build on arm64 to confirm it works.


jesseduffield:
Thanks for the PR but this is now obsoleted by https://github.com/jesseduffield/lazydocker/pull/561

Closing

PR #549, opened 2024-06-16
Feature/ctrl l clear logs

peauc:
Following work from #403


peauc:
The keybind seems to be correctly configured but it doesn't seem to trigger on my machine. Looking for feedback :eyes:

fsmanuel:
Any progress? Would love to use this feature!

cc @jesseduffield

ahmadie:
clear the logs!! need this thumb up only

PR #551, opened 2024-06-18
Fix #512

### LINKED ISSUE #512: XDG base directory specification is not described in the config docs
After #129 is merged the [docs](https://github.com/jesseduffield/lazygit/blob/master/docs/Config.md) is still old and does not mention `XDG_CONFIG_HOME`.
It is better to be documented like [the docs of lazygit](https://github.com/jesseduffield/lazygit/blob/master/docs/Config.md).

peauc:
Add the fact that lazydocker uses XDG Spec for the config file

### DIFF
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -4,13 +4,16 @@
 
 The location of the user config will differ depending on your OS. You can open it via lazydocker by opening the application, clicking on the 'project' panel at the top left and pressing 'o' (or pressing 'e' if your files open in vim).
 
-Changes to the user config will only take place after closing and re-opening lazydocker
+Changes to the user config will only take place after closing and re-opening lazydocker.
 
 ### Locations:
 
 - OSX: `~/Library/Application Support/jesseduffield/lazydocker/config.yml`
 - Linux: `~/.config/lazydocker/config.yml`
 - Windows: `C:\\Users\\<User>\\AppData\\Roaming\\jesseduffield\\lazydocker\\config.yml` (I think)
+- UNIX when `XDG_CONFIG_HOME` is found: `$XDG_CONFIG_HOME/lazydocker/config.yml`
+
+Lazydocker follows the [XDG base directory specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) meaning that the config will be placed in `XDG_CONFIG_HOME` if the variable is found.
 
 JSON schema is available for `config.yml` so that IntelliSense in Visual Studio Code
 (completion and error checking) is automatically enabled when the [YAML Red Hat][yaml]


peauc:
Fix #512

docs/Config.md
```diff
@@ -4,13 +4,16 @@
...
 
-Changes to the user config will only take place after closing and re-opening lazydocker
+Changes to the user config will only take place after closing and re-opening lazydocker.
 
 ### Locations:
 
 - OSX: `~/Library/Application Support/jesseduffield/lazydocker/config.yml`
 - Linux: `~/.config/lazydocker/config.yml`
 - Windows: `C:\\Users\\<User>\\AppData\\Roaming\\jesseduffield\\lazydocker\\config.yml` (I think)
+- UNIX when `XDG_CONFIG_HOME` is found: `$XDG_CONFIG_HOME/lazydocker/config.yml`
```
jesseduffield, line 14: Readers may stop reading once they see OSX and Linux. We should say on each of those lines '(if XDF_CONFIG_HOME env variable is not set)' and we should have the unix line first.

jesseduffield (changes requested):
One comment

docs/Config.md
peauc: Alright i've changed the wording, let me know what you think

peauc:
I've also squashed commits on this PR

docs/Config.md
```diff
@@ -4,14 +4,18 @@
...
 
 ### Locations:
 
-- OSX: `~/Library/Application Support/jesseduffield/lazydocker/config.yml`
-- Linux: `~/.config/lazydocker/config.yml`
+Lazydocker follows the [XDG base directory specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) meaning that the config will be placed in `XDG_CONFIG_HOME` if the variable is found.
+
+- UNIX(if `XDG_CONFIG_HOME` environment variable is found): `$XDG_CONFIG_HOME/lazydocker/config.yml`
+- OSX(if `XDG_CONFIG_HOME` environment variable is **not** found): `~/Library/Application Support/jesseduffield/lazydocker/config.yml`
+- Linux(if `XDG_CONFIG_HOME` environment variable is **not** found): `~/.config/lazydocker/config.yml`
```
sequencerr, line 15: Imo, this should take place as a note. after. even if readers "stop reading", instead of bloating.

docs/Config.md
```diff
@@ -4,14 +4,18 @@
...
 The location of the user config will differ depending on your OS. You can open it via lazydocker by opening the application, clicking on the 'project' panel at the top left and pressing 'o' (or pressing 'e' if your files open in vim).
 
-Changes to the user config will only take place after closing and re-opening lazydocker
+Changes to the user config will only take place after closing and re-opening lazydocker.
 
 ### Locations:
 
-- OSX: `~/Library/Application Support/jesseduffield/lazydocker/config.yml`
-- Linux: `~/.config/lazydocker/config.yml`
+Lazydocker follows the [XDG base directory specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) meaning that the config will be placed in `XDG_CONFIG_HOME` if the variable is found.
```
sequencerr, line 11: maybe rephrase?
`Lazydocker adheres to the XDG Base Directory Specification, placing its configuration in XDG_CONFIG_HOME if the variable is set.`

docs/Config.md
jesseduffield: I agree: this should just say that if the XDG_CONFIG_HOME env var is set, we use $XDG_CONFIG_HOME/lazydocker/config.yml, then the rest can be stated as fallbacks without being as verbose

PR #554, opened 2024-06-26
Add hint to what .DockerCompose is in documentation

hauskens:
It took me a fair bit of time before i understood that '{{ .DockerCompose }}' was defined by the dockerCompose var in commandTemplates.

In hindsight it is VERY obvious.. so i just wanted to contribute a comment that would have been useful to me when i first tried to set this up.


jesseduffield:
LGTM

PR #576, opened 2024-08-29
Added configuration options to hide side panels

### LINKED ISSUE #513: Config options for side panel visibility / size
I have recently updated from an older version and the "network" side panel has since been added (https://github.com/jesseduffield/lazydocker/pull/424).
I work on a quite small screen and am really only interested in the services, containers and images side panels.
In the past, the volumes side panel already "took away" some vertical space and now the networks section takes away even more.

Option to hide specific side panels via the config file.

- Make the panel height configurable so that i can "clamp" the height of the network and volume panel to something very small and leave more room for the others.
- A more extreme version of the accordion effect (expandFocusedSidePanel) that makes unfocused side panels even smaller.

jayrhynas:
Every panel except "project" can be hidden. All panels are visible by default.
I decided to not let "project" be hidden since a) it's special cased in various layouts and b) it doesn't really make sense to have *no* panels visible at all.

`initiallyFocusedViewName` defaults to "project" if "services" and "containers" are hidden

This should fix #513.

### DIFF
--- a/docs/Config.md
+++ b/docs/Config.md
@@ -48,6 +48,13 @@ gui:
   wrapMainPanel: true
   # Side panel width as a ratio of the screen's width
   sidePanelWidth: 0.333
+  # Determines whether each side panel is shown or not
+  sidePanelVisibility:
+    services: true
+    containers: true
+    images: true
+    volumes: true
+    networks: true
   # Determines whether we show the bottom line (the one containing keybinding
   # info and the status of the app).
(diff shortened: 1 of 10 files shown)


madsaylor:
Really looking forward to have it

yurighensev:
Any chances this would move forward? Would be a great addition.

magnus-larsson:
Any updates regarding this update?

Issue #559, opened 2024-06-28
Docker context using SSH does not work


peauc:
**Describe the bug**
I have a docker context refering to `ssh://root@peauc-homeserver` lazydocker fails to connect
**To Reproduce**
As stated above create a docker context to a remote computer running to which you can connect using ssh wihout password on the key. It will fail
**Expected behaviour**
It should work

**Screenshots**
![CleanShot 2024-06-28 at 22 36 45@2x](https://github.com/jesseduffield/lazydocker/assets/15132860/2c658022-0313-4092-a184-5dd2b4c62759)

**Desktop (please complete the following information):**
 - OS: OSx
 - Lazydocker Version: Latest

PR #511, opened 2024-02-17
fix: Properly support remote docker daemon over ssh

rajiv-k:
This issue looks like a possible regression. This PR fixes the issue with connecting to remote docker daemon over SSH.

Fixes #510

PR #577, opened 2024-09-02
Fix ssh context not working

peauc:
Fix issue as described in #559

### DIFF
--- a/pkg/commands/docker.go
+++ b/pkg/commands/docker.go
@@ -69,19 +69,29 @@ func (c *DockerCommand) NewCommandObject(obj CommandObject) CommandObject {
 	return defaultObj
 }
 
-// NewDockerCommand it runs docker commands
+// NewDockerCommand creates a DockerCommand struct that will be able to run docker commands.
 func NewDockerCommand(log *logrus.Entry, osCommand *OSCommand, tr *i18n.TranslationSet, config *config.AppConfig, errorChan chan error) (*DockerCommand, error) {
-	tunnelCloser, err := ssh.NewSSHHandler(osCommand).HandleSSHDockerHost()
-	if err != nil {
-		ogLog.Fatal(err)
-	}
-
 	dockerHost, err := determineDockerHost()
 	if err != nil {
 		ogLog.Printf("> could not determine host %v", err)
 	}
 
-	cli, err := client.NewClientWithOpts(client.FromEnv, client.WithVersion(APIVersion), client.WithHost(dockerHost))
+	tunnelCloser, err := ssh.NewSSHHandler(osCommand).HandleSSHDockerHost(dockerHost)
+	if err != nil {
+		ogLog.Fatal(err)
+	}
+
+	clientOpts := []client.Opt{
+		client.FromEnv,
+		client.WithVersion(APIVersion),
+	}
+	// For an ssh connection the DOCKER_HOST env variable has been overridden.
+	// Discard the previously determined dockerHost
+	if !strings.HasPrefix(dockerHost, "ssh://") {
+		clientOpts = append(clientOpts, client.WithHost(dockerHost))
+	}
+
+	cli, err := client.NewClientWithOpts(clientOpts...)
 	if err != nil {
 		ogLog.Fatal(err)
 	}
--- a/pkg/commands/ssh/ssh.go
+++ b/pkg/commands/ssh/ssh.go
@@ -44,10 +44,11 @@ func NewSSHHandler(oSCommand CmdKiller) *SSHHandler {
 
 // HandleSSHDockerHost overrides the DOCKER_HOST environment variable
 // to point towards a local unix socket tunneled over SSH to the specified ssh host.
-func (self *SSHHandler) HandleSSHDockerHost() (io.Closer, error) {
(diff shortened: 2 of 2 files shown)


peauc:
It seems that I was missing the code of the PR #511 that was merged early June.
I still have the issue using my now fresh copy of master.

The code of this branch is a bit cleaner and now works for me

peauc:
I'll get to the tests in a few days.

pkg/commands/docker.go
```diff
@@ -71,33 +71,30 @@ func (c *DockerCommand) NewCommandObject(obj CommandObject) CommandObject {
 	return defaultObj
 }
 
-// NewDockerCommand it runs docker commands
+// NewDockerCommand creates a DockerCommand struct that wraps the docker client.
+// Able to run docker commands. And handles
```
jesseduffield, line 75: This comment is incomplete

pkg/commands/docker.go
```diff
@@ -71,33 +71,30 @@ func (c *DockerCommand) NewCommandObject(obj CommandObject) CommandObject {
...
-	// Retrieve the docker host from the environment which could have been set by
-	// the `SSHHandler.HandleSSHDockerHost()` and override `dockerHost`.
-	dockerHostFromEnv := os.Getenv(dockerHostEnvKey)
-	if dockerHostFromEnv != "" {
-		dockerHost = dockerHostFromEnv
+	clientOpts := []client.Opt{
+		client.FromEnv,
+		client.WithVersion(APIVersion),
+	}
+	// For an ssh connection the DOCKER_HOST env variable has been overridden.
```
jesseduffield, line 91: Could you explain the rationale behind these changes? I'm not clear on the reasoning

jesseduffield (changes requested):
See comments :)

pkg/commands/docker.go
peauc: I just wired it correctly for the SSH connection to work.
I'll rework it so its clearer and doesn't rely on a modified docker host env variable.

pkg/commands/docker.go
peauc: Very correct. Thanks pointing it out.

pkg/commands/docker.go
peauc: Hey @jesseduffield took over 6 months to get some free time but now that I'm in vacation... :D

The rationale is that ATM ssh tunneling with docker context over lazydocker is broken. There are a few issues regarding this.
In my opinion those issues are due to an edge case in the current implementation of the ssh tunneling & docker host parsing.
I've rewritten a few parts to both make it simpler and more robust, tell me what you think :)

bill-mccormick-dg:
Anyway this get's merged soon @jesseduffield? I'd love to have it.

wpcarro:
What's the latest here?

peauc:
@wpcarro waiting for merge on this repo. Works on my fork.

bill-mccormick-dg:
Sorry haven't tried this. I'm on a work break and not sure I'll have a
chance to look at it until next week. BTW...my use case is from WSL.

peauc:
If anyone is interested in this thread. I've released `lazydocker-ng` with a bunch of fixes that where never merged.

francoiscabrol:
I am facing the same issue. Thank you, @peauc, for providing a solution. Would it be possible to review it again, @jesseduffield? I have checked with the fork, and the issue is gone. It would be awesome to have it merge in the main project!

francoiscabrol:
I have an update: maybe my issue was a bit different. My issue was about getting `ssh tunneled socket never became available: context deadline exceeded` (like https://github.com/jesseduffield/lazydocker/issues/591) running `DOCKER_HOST=ssh://user@domain:2222 ~/go/bin/lazydocker`
With the fork `lazydocker-ng`, this command works without error.

But I realize that I can actually connect through ssh with the main project if I omit the port in the command `DOCKER_HOST=ssh://user@domain ~/go/bin/lazydocker`

peauc:
@francoiscabrol, yeah there is an issue with the way the docker host is consumed. 
Your workaround is a good one if you have the default ssh port.

vlanx:
> I have an update: maybe my issue was a bit different. My issue was about getting `ssh tunneled socket never became available: context deadline exceeded` (like #591) running `DOCKER_HOST=ssh://user@domain:2222 ~/go/bin/lazydocker` With the fork `lazydocker-ng`, this command works without error.
> 
> But I realize that I can actually connect through ssh with the main project if I omit the port in the command `DOCKER_HOST=ssh://user@domain ~/go/bin/lazydocker`

This did not work for me either, port or no port. 
What did though, was creating an entry in my `ssh` config file like so:
```
Host remote-context
  Hostname 10.16.10.1
  User ubuntu
```
Then, the `DOCKER_HOST` url can be `DOCKER_HOST=ssh://remote-context lazydocker` which makes it work for me.
I've wrapped into a `zsh` helper function and autocompletes with `ssh` aliases too.
```
lzd() {
  local host_alias="$1"

  if [ -z "$host_alias" ]; then
    echo "usage: lzd <ssh-host-alias>"
    return 1
  fi

  DOCKER_HOST="ssh://$host_alias" lazydocker
}

compdef lzd=ssh
```

PR #584, opened 2024-10-14
Add clipboard support & feature to copy container id

### LINKED ISSUE #580: Copy container id to clipboard
Loving lazydocker, however often have to run a `docker container ps`, in order to copy a container id or name in order to do a `docker cp file.txt containerid:/some/path/file.txt`.

In the lazydocker tui, hovering over a running container, a key binding should copy container id to the system clipboard.

peauc:
Fix #580

### DIFF
--- a/pkg/gui/app_status_manager.go
+++ b/pkg/gui/app_status_manager.go
@@ -1,6 +1,7 @@
 package gui
 
 import (
+	"sync"
 	"time"
 
 	"github.com/jesseduffield/gocui"
@@ -15,10 +16,15 @@ type appStatus struct {
 
 type statusManager struct {
 	statuses []appStatus
+	lock     *sync.Mutex
 }
 
 func (m *statusManager) removeStatus(name string) {
 	newStatuses := []appStatus{}
+
+	m.lock.Lock()
+	defer m.lock.Unlock()
+
 	for _, status := range m.statuses {
 		if status.name != name {
 			newStatuses = append(newStatuses, status)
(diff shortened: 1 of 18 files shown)


pkg/gui/app_status_manager.go
```diff
@@ -28,6 +34,9 @@ func (m *statusManager) removeStatus(name string) {
 }
 
 func (m *statusManager) addWaitingStatus(name string) {
+	m.lock.Lock()
+	defer m.lock.Unlock()
+
 	m.removeStatus(name)
```
jesseduffield, line 40: Given that removeStatus tries to obtain the lock that addWaitingStatus already holds, wouldn't that cause a deadlock?

pkg/gui/containers_panel.go
```diff
@@ -358,6 +358,20 @@ func (gui *Gui) PauseContainer(container *commands.Container) error {
 	})
 }
 
+func (gui *Gui) handleCopyContainerId(g *gocui.Gui, v *gocui.View) error {
+	ctr, err := gui.Panels.Containers.GetSelectedItem()
+	if err != nil {
+		return nil
+	}
+
+	err = gui.WithStaticWaitingStatus(fmt.Sprintf(gui.Tr.CopyContainerIdStatus, utils.TruncateWithEllipsis(ctr.ID, 10)), time.Second*2)
```
jesseduffield, line 367: This isn't really a waiting status: it's a toast message. We should update the naming to reflect that (and I haven't tested it locally but it's important we don't show a loading spinner on this)

jesseduffield (commented):
Sorry for late review: I'm going to try and chip away at this stuff over time. Couple comments

pkg/gui/containers_panel.go
peauc: yes thanks for the feedback. I'll go over it again. 👍

pkg/gui/app_status_manager.go
peauc: There might be issues with my understanding of the locks. I'll double/triple check 👍

PR #594, opened 2024-11-19
Name of container should be checked first

umermala:
In most of the cases, the base image is used without making any changes and name label is not provided however, in most use cases, name is provided therefore the Name key should be checked first and then label name and if none is given then just default name

### DIFF
--- a/pkg/commands/docker.go
+++ b/pkg/commands/docker.go
@@ -259,7 +259,7 @@ func (c *DockerCommand) GetContainers(existingContainers []*Container) ([]*Conta
 		} else if name, ok := ctr.Labels["name"]; ok {
 			newContainer.Name = name
 		} else {
-			newContainer.Name = "default_name"
+			newContainer.Name = "Unknown"
 		}
 		newContainer.ServiceName = ctr.Labels["com.docker.compose.service"]
 		newContainer.ProjectName = ctr.Labels["com.docker.compose.project"]


hamzanajeebkhan (approved):

pkg/commands/docker.go
```diff
@@ -253,11 +253,13 @@ func (c *DockerCommand) GetContainers(existingContainers []*Container) ([]*Conta
...
-		// if the container is made with a name label we will use that
-		if name, ok := ctr.Labels["name"]; ok {
+		// if the container is made with a name then use that directly
+		if len(ctr.Names) > 0 && ctr.Names[0] != "" {
+			newContainer.Name = strings.TrimLeft(ctr.Names[0], "/")
+		} else if name, ok := ctr.Labels["name"]; ok {
 			newContainer.Name = name
 		} else {
-			newContainer.Name = strings.TrimLeft(ctr.Names[0], "/")
+			newContainer.Name = "default_name"
```
jesseduffield, line 262: I'd call this 'unknown'

jesseduffield (changes requested):

pkg/commands/docker.go
umermala: Sure lets call it unknown

umermala: Please review now

Issue #597, opened 2024-11-25
Jump to panel keybinding


A11Might:
**Is your feature request related to a problem? Please describe.**
No.

**Describe the solution you'd like**
Use the bound keys to select the specified panel, just like lazygit.

**Describe alternatives you've considered**
No.

**Additional context**
No.

A11Might:
same issue: https://github.com/jesseduffield/lazydocker/issues/522

PR #540, opened 2024-06-05
Add navigation with numbers

peauc:
Fix #522 
Like lazygit we can navigate with numbers.

PR #598, opened 2024-11-25
feat: use bound keys select side panel

A11Might:
This PR add feature https://github.com/jesseduffield/lazydocker/issues/597

Set the showPanelJumps configuration to true, and use the assigned keys to quickly navigate and select specific panels, similar to how it's done in LazyGit.


A11Might:
Sorry, just saw this PR：https://github.com/jesseduffield/lazydocker/pull/540

PR #616, opened 2025-01-10
Fix bug `No such file or directory` in  install_update_linux.sh

zcorn2017:
Create the directory if it does not exist.
By this fix, we can avoid the error:
`install: failed to access ‘~/.local/bin’: No such file or directory`

### DIFF
--- a/scripts/install_update_linux.sh
+++ b/scripts/install_update_linux.sh
@@ -3,6 +3,9 @@
 # allow specifying different destination directory
 DIR="${DIR:-"$HOME/.local/bin"}"
 
+# ensure the directory exists
+mkdir -p "$DIR"
+
 # map different architecture variations to the available binaries
 ARCH=$(uname -m)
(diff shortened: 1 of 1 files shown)


mark2185:
That's odd, `install -D` should suffice. We had the same issue on [lazygit](https://github.com/jesseduffield/lazygit/pull/4031), but `-D` is present here already.

zcorn2017:
This issue happens when the `.config` folder does not exist, which is typical after a fresh install.
`install -D` only creates the folder when the parent folder `.config` exists.

zcorn2017:
> That's odd, `install -D` should suffice. We had the same issue on [lazygit](https://github.com/jesseduffield/lazygit/pull/4031), but `-D` is present here already.

This issue happens when the `.config` folder does not exist, which is typical after a fresh install.
`install -D` only creates the folder when the parent folder `.config` exists.

PR #630, opened 2025-02-12
Add project name to docker command

Groysman881:
Determing docker compose project name provides the ability to work with services of each docker compose project separately in various Lazydocker instances, even if services have the same names

Determing project name is only available if name explicitly specified in docker compose project config (docker compose v2)


epicwhale:
+1 to having richer project name support - there's a bunch of erratic behaviors when using `-f` to launch a docker compose project that spans across multiple files, and then trying to restart containers of the project from inside lzd. I don't think the commands it generates by default are scoped to the same compose project

PR #633, opened 2025-03-06
fix: change the service -> container mapping to map by id

jotavemonte:
Change the approach to service -> container mapping. The previous approach used the service name to connected them. The new approach maps the service to the container by id.

I tested it with my personal conflicting projects and it managed to fix it.

It fixes: #122


nicolae-gorgias:
nice work!

nicolae-gorgias:
no review yet? don't be discouraged 💪

PR #637, opened 2025-03-24
docs: add lazydocker installation script for Ubuntu

rejmann:
Adds script to install the latest version of lazydocker on Ubuntu.
Tested on Ubuntu 20.04 and 22.04.


Emkas:
Well, I don't think it is needed. Basically this is the same as binary release, but installs in `usr/local/bin` not in `$HOME/.local/bin`, but there is a `DIR` variable for that.

rejmann:
No prob @Emkas

PR #638, opened 2025-03-28
Support for docker profiles

### LINKED ISSUE #423: profile support
I would be happy if docker-compose profiles are supported in `lazydocker`.

saravanabalagi:
Support for docker profiles
- Project panel also shows profiles available
- Up and down profiles in project panel optionally with volumes
- Up and down profiles in services panel as well

Closes #423 

---

Why is it in project and not in a separate panel?
- Profiles behave like project, upping profile starts all base services (without any profiles assigned) and any service that has the profile name listed under profiles key
- In one sense, the project itself is a profile
- So it made sense to me to have this in the project panel

What about keybindings?
- `u` and `d` ups and downs profiles, very much like how these ups/downs a service in services panel
- `U` and `D` shows menu to up and down project/profile, very much like how they behave in services panel

I made these choices because it felt intuitive and found it comfortable during my daily driving the same.

---

Screenshots:

1. Project panel:
image

2. Project panel menu:
image

3. Services panel menu:
image

---

FAQs:

1. I upped the base services by upping the project, then upped services for `profile_n`. But when down `profile_n`, it also shuts down base services?

Yes, that's because upping `profile_n` will also start base services. And downing will shut down base services as well. This is how `--profile` works in docker CLI.

2. What is needed if I want `d` to down only services pertaining to `profile_n`?

It felt to me that this would be intuitive than shutting down base services additionally. But unfortunately, AFAIK in order to achieve this behavior, we need to find a list of all services and filter the ones that belong to the profile and then run down individually on each service.


saravanabalagi:
Heads up, I also updated missing translations, so the PR will show too many changes.

saravanabalagi:
Adding binaries for testing:

[lazydocker-linux-arm64.tar.gz](https://github.com/user-attachments/files/19636788/lazydocker-linux-arm64.tar.gz)
[lazydocker-linux-amd64.tar.gz](https://github.com/user-attachments/files/19636792/lazydocker-linux-amd64.tar.gz)
[lazydocker-darwin-arm64.tar.gz](https://github.com/user-attachments/files/19636793/lazydocker-darwin-arm64.tar.gz)
[lazydocker-darwin-amd64.tar.gz](https://github.com/user-attachments/files/19636796/lazydocker-darwin-amd64.tar.gz)

PR #650, opened 2025-05-04
Update deprecated actions dependencies

peauc:
We are using older versions of actions that are currently being deprecated and have brownout periods where the CDN doesn't answer for those versions.

https://github.com/actions/cache/discussions/1510
https://github.com/orgs/community/discussions/151729

Bump to latest actions version. Should be retro compatible.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -10,15 +10,15 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout code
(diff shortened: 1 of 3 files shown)


peauc:
CleanShot 2025-05-04 at 14 24 29@2x
Still some issues with linting config. Will check

peauc:
CleanShot 2025-05-04 at 14 31 27@2x
Returns OK. We should upgrade our go version though.

Issue #635, opened 2025-03-19
Feature Request: Order `Env` alphabetically


thi-marques:
It would be easier to locate a environment variable if they were ordered alphabetically.

![Image](https://github.com/user-attachments/assets/ee52f03f-f0e3-4c2a-8e9a-a3a2984d0604)

Maybe it could be a option in the config file.

volodimyr:
Done

etiedt:
Also would really like to see this. Long lists of environment variables become hard to parse when they aren't sorted.

PR #653, opened 2025-05-31
Order environment variables alphabetically

volodimyr:
#635

### DIFF
--- a/pkg/gui/containers_panel.go
+++ b/pkg/gui/containers_panel.go
@@ -3,6 +3,7 @@ package gui
 import (
 	"context"
 	"fmt"
+	"sort"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


pkg/gui/containers_panel.go
```diff
@@ -135,7 +136,10 @@ func (gui *Gui) containerEnv(container *commands.Container) string {
 		return gui.Tr.NothingToDisplay
 	}
 
-	envVarsList := lo.Map(container.Details.Config.Env, func(envVar string, _ int) []string {
+	envVars := container.Details.Config.Env
```
manascb1344: envVars := container.Details.Config.Env creates a reference to the same underlying array.
sort.Strings(envVars) sorts in place, so it also sorts container.Details.Config.Env.
This mutates the original data, which is risky even if it's only used for display.

As the code mutates the original slice. Please create a copy before sorting

PR #673, opened 2025-08-20
Add memory used and limit to stat

### LINKED ISSUE #665: feat: show memory usage in MB/GB (not just percent) in stats tab
I'm often frustrated when trying to assess container memory usage in Lazydocker, because the Stats tab only shows memory usage as a percentage. This makes it unclear how much RAM a container is actually using, especially when there’s no memory limit set or the total system memory is not immediately obvious.

This becomes even more problematic when comparing containers running across different machines with varying amounts of RAM, where the percentage loses context and makes it difficult to compare usage meaningfully.

I’d like Lazydocker to display memory usage in MB or GB, either alongside or instead of the percentage in the `Stats` tab. For example:
`Memory: 125 MB / 512 MB (24.4%)`

dimonl:
closes [#665](https://github.com/jesseduffield/lazydocker/issues/665)


yugui923:
It would be wonderful to see these stats too. Thank you.

PR #690, opened 2025-10-13
Scroll by half page on page up/down

### LINKED ISSUE #267: Page up/down keys only scroll one line, home does not jump to top.
**Describe the bug**
In a log or configuration view, pressing the page up or down buttons do not scroll a page. 

Also while the "End" button jumps to the bottom, the "Home" button does not jump to the top.

**To Reproduce**
1. Go to container config
2. Press page down key.
3. See config scroll 1 line.

**Expected behavior**
Config readout should scroll my one page worth of lines.
Home button should jump from current position to the top.

 - OS: PopOS 21.04
 - Lazydocker Version 0.12

brdoney:
Took a stab at making page up/down and ctrl u/d scroll by half a page, rather than just a single line.

Fixes #267


rdlmda:
I'd love to see it implemented!

Issue #698, opened 2025-11-07
Error - Incorrect client version detection


PaPaTheGMan:
**Describe the bug**
When running lazydocker an error message reports:
Docker eventstream returned error: Error response from daemon: client version 1.25 is too old. Minimum supported API version is 1.44, please upgrade your client to a newer version.

**To Reproduce**
1. enter 'lazydocker' from any terminal screen

**Expected behaviour**
Lazydocker start up and list active containers, etc. with no error

**Screenshots**

<img width="1094" height="579" alt="Image" src="https://github.com/user-attachments/assets/cd853c9b-6c86-4668-becb-1ae091a3a869" />

**Desktop (please complete the following information):**
 - OS: Ubuntu 25.10 x86_64
 - Lazydocker Version 24.1

**Additional context**
Docker Version Info: (note API is 1.52, well above 1.44 minimum)
Client: Docker Engine - Community
 Version:           29.0.0-rc.3
 API version:       1.52
 Go version:        go1.25.3
 Git commit:        ee244f2
 Built:             Fri Nov  7 00:41:58 2025
 OS/Arch:           linux/amd64
 Context:           default

Server: Docker Engine - Community
 Engine:
  Version:          29.0.0-rc.3
  API version:      1.52 (minimum version 1.44)
  Go version:       go1.25.3
  Git commit:       cb8228b
  Built:            Fri Nov  7 00:41:58 2025
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          v1.7.29
  GitCommit:        442cb34bda9a6a0fed82a2ca7cade05c5c749582
 runc:
  Version:          1.3.3
  GitCommit:        v1.3.3-0-gd842d771
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0

Nekomich:
I can confirm that this also occurs under Debian.

_Error response from daemon: client version 1.25 is too old. Minimum supported API version is 1.44, please upgrade your client to a newer version_
I get the message by mail from Watchtower.


**System:**
Debian GNU/Linux 13 (trixie)
latest Watchtower release

**Additional context:**
Client: Docker Engine - Community
 Version:           29.0.0
 API version:       1.52
 Go version:        go1.25.4
 Git commit:        3d4129b
 Built:             Mon Nov 10 21:46:28 2025
 OS/Arch:           linux/amd64
 Context:           default

Server: Docker Engine - Community
 Engine:
  Version:          29.0.0
[...]

PR #699, opened 2025-11-11
Error - Incorrect client version detection #698

FabianRingel:
changed Docker API version to 1.44

Just changing the variable forks fine for me.
tested on debian 13

### DIFF
--- a/pkg/commands/docker.go
+++ b/pkg/commands/docker.go
@@ -28,7 +28,7 @@ import (
 )
 
 const (
-	APIVersion       = "1.25"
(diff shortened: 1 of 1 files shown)


zyncc:
please approve this pr asap

obouchet:
in the meantime, [oxker](https://github.com/mrjackwills/oxker) could be a backup solution. It's not perfect and it's missing images management but at least it's not having the docker client compatibility issue.

jesseduffield:
went with this instead https://github.com/jesseduffield/lazydocker/pull/703

PR #708, opened 2025-11-16
feat: implement new languages: Italian, Russia, Japanese

EgorYolkin:
Add new localization support for Italian, Russian and Japanese
in pkg/i18n.

All three languages are wired into the existing i18n loader and
are available through the same translation lookup mechanism.


natsuki-engr:
@jesseduffield 
Gentle ping on this PR in case it got buried. I would love to see this merged.

PR #732, opened 2025-12-31
Container logs filtering

### LINKED ISSUE #655: Filter Container Log Output
Hi, I have a feature request for lazydocker that I am willing to implement, if the owners approve of it. 

Whenever I am viewing log output for container, I frequently desire a way to search or filter for a particular keyword in the output. I know of no current support for this feature, and so I always revert to exiting lazydocker, and grepping the output of `docker logs`.

I think this would be particularly useful for devs like me: I often have a dozen or more containers running locally, and I to debug issues with our microservices ,I need to search for keywords across 3 or more of those containers. 

To maintain consistency with the filter that already exists in the `Services` pane, I suggest the `/` keyboard shortcut be used. When the `Logs` pane is focused, pressing `/` would enable the `filter:` input at the bottom of the screen. As you type, the output would be filtered in real time for any lines that contain that pattern. You should be able to press enter to commit to that filter, just like pressing enter for the `Services` pane filter, and escape to remove that filter.

- `docker logs <container name> | grep <pattern>`
- `docker compose logs | grep <pattern>`

Having to break out and run `docker logs` on each of those containers is tedious. Running docker logs on the entire compose service cluster can be slow or noisy. I also just like lazydocker and prefer to stay in its context when I'm already there.

ManManavadaria:
## Summery
This PR adds the ability to filter on container logs by pressing the key '/' while the Logs tab is in focus

## Motivation
I think this feature is practically useful because while working on containers with large volume of logs, it can be difficult to quickly find relevant entries. This feature enables filtering without changing context, eliminating the need to scroll through large log streams or switch scope to run docker logs | grep commands

Closes #655 

## Changes
- Added real-time filtering for container logs (press `/` while viewing logs)
- Added LogsFilter state management
- Implemented streaming log filtering for both TTY and non-TTY logs
- Refactored filtering to support both panel list and logs filtering

image

### DIFF
--- a/docs/keybindings/Keybindings_de.md
+++ b/docs/keybindings/Keybindings_de.md
@@ -97,6 +97,7 @@ _This file is auto-generated. To update, make the changes in the pkg/i18n direct
 
 <pre>
   <kbd>esc</kbd>: zurück
+  <kbd>/</kbd>: filter logs
 </pre>
 
 ## Global
--- a/docs/keybindings/Keybindings_en.md
+++ b/docs/keybindings/Keybindings_en.md
@@ -97,6 +97,7 @@ _This file is auto-generated. To update, make the changes in the pkg/i18n direct
 
 <pre>
   <kbd>esc</kbd>: return
+  <kbd>/</kbd>: filter logs
 </pre>
 
 ## Global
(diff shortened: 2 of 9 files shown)


ManManavadaria:
Hi @jesseduffield 

I opened this PR a while ago to add log filtering feature. I just wanted to gently follow up.

Whenever you have a chance, I’d really appreciate it if you could take a look and share your thoughts. If the approach or feature isn’t aligned with the project’s direction, I’d be happy to revise it or close the PR accordingly.

Thanks for your time!

Issue #748, opened 2026-01-27
add force remove variant


gajus:
I would like there to be an option to force remove containers without needing to first stop them. 

<img width="439" height="109" alt="Image" src="https://github.com/user-attachments/assets/479f933f-e8fe-4343-a124-2da8e2735fdc" />

achintya-7:
Big +1 for this; it would streamline things significantly. This is a major friction point in my own workflow, so I’m going to go ahead and implement it.

My plan is to update the Remove method in `pkg/commands/docker_commands/container.go` to support the force flag and then expose the new options in the container action menu. I'll submit a PR shortly—let me know if you have any specific preferences for the menu layout!

achintya-7:
I have added a PR for this. Let's wait for it to get reviewed.

PR #751, opened 2026-02-08
Force Remove Option for Containers

achintya-7:
### Summary
Adds two options to the container remove menu so users can remove running containers without stopping them first (equivalent to docker rm --force).
Fixes/Implements: #748

### Changes
* Force remove – `docker rm --force <container>`. Uses existing gui.Tr.ForceRemove. Calls Container.Remove with RemoveOptions{Force: true}.
* Force remove with volumes – `docker rm --force --volumes <container>`. Uses gui.Tr.ForceRemoveWithVolumes. Calls Container.Remove with RemoveOptions{Force: true, RemoveVolumes: true}.

### i18n
* Added ForceRemoveWithVolumes to the translation set and to all language files in pkg/i18n/ for the second menu label.


achintya-7:
@jesseduffield pinging once :eyes:

PR #773, opened 2026-03-06
fix: show containers in network config view

### LINKED ISSUE #752: When selecting a network, "Containers" in config shows "none"
**Describe the bug**
When I select a network, the Config on the right shows "Containers: none". However, if I run "docker networks inspect NETWORK_NAME", I see many containers in that network.

**To Reproduce**
1. Select a network on the left panel
2. View "Containers" on the right panel.

**Expected behaviour**
I expect a list of containers that are in that network, similar to how it displays "Labels"

 - OS: Arch Linux
 - Lazydocker Version 0.24.4

umut-polat:
When selecting a network, the config panel always shows `Containers: none` even when containers are attached to it.

The `NetworkList` API doesn't populate the `Containers` field — only `NetworkInspect` does. The config renderer was reading from the list response, which is always empty.

Added a `Network.Inspect()` method and call it when rendering the config panel to fetch the full network details including connected containers.

Fixes #752

### DIFF
--- a/pkg/commands/network.go
+++ b/pkg/commands/network.go
@@ -48,6 +48,11 @@ func (c *DockerCommand) PruneNetworks() error {
 	return err
 }
 
+// Inspect fetches the full network details including containers
+func (v *Network) Inspect() (network.Inspect, error) {
+	return v.Client.NetworkInspect(context.Background(), v.Network.ID, network.InspectOptions{})
+}
+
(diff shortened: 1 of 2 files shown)


umut-polat:
friendly ping — this is still clean and ready for review. happy to address any feedback. thanks!

PR #813, opened 2026-06-18
docs: list all supported languages for gui.language

s3onghyun:
Config.md documents `gui.language` as one of 'auto' | 'en' | 'pl' | 'nl' | 'de' | 'tr', but pkg/i18n registers four more (fr, zh, es, pt) that already ship with translation files. Added them so the documented set matches the code.


s3onghyun:
Tidying up my open PRs — closing this minor docs fix. It's a trivial one-liner, so no loss if it isn't worth a review cycle; feel free to cherry-pick the change if it's ever useful. Thanks for maintaining the project!

PR #814, opened 2026-06-18
Add Korean (ko) translation

s3onghyun:
## Add Korean (ko) translation

lazydocker already ships nine locales (en, pl, nl, de, tr, fr, zh, es, pt) but not Korean. This adds a Korean translation set and registers the `ko` language code in `GetTranslationSets()`, so Korean users get a localized UI when their `LANG` resolves to Korean (or via `gui.language: ko`).

Like the existing locales, untranslated keys fall back to English through the existing `mergo.Merge` in `NewTranslationSet`, so nothing breaks if a string is missing.

I'm a native Korean speaker; the wording follows the same casual/imperative tone the other locales use. Technical terms commonly left in English by Korean developers (Docker, compose, attach) are kept recognizable.

Verified locally with `GOFLAGS=-mod=vendor`: `gofmt -l` clean, `go build ./...` and `go vet ./pkg/i18n/` pass.


s3onghyun:
Closing this to refocus my open-source time on a smaller set of projects. The change still stands on its own — happy to reopen if a maintainer would like to pick it up. Thanks for maintaining this!

PR #830, opened 2026-08-08
Add SSH keepalives to the Docker host tunnel

### LINKED ISSUE #815: EOF error when streaming Docker events over SSH context
**Describe the bug**

Lazydocker fails when connected to a remote Docker Engine via a Docker SSH context. The Docker CLI works correctly (`docker ps`, `docker events`, etc.), but Lazydocker reports an error when subscribing to the Docker events stream.

Error displayed:

```text
Docker event stream returned error:
error during connect:
Get "http://%2Ftmp%2Flazydocker-sshtunnel-XXXXXXXX/dockerhost.sock/v1.51/events":
EOF
```

The container list loads successfully, but the event stream disconnects shortly afterward.

**To Reproduce**

1. Create a Docker context that connects to a remote Docker host over SSH:

   ```bash
   docker context create AORUS --docker "host=ssh://AORUS"
   docker context use AORUS
   ```

2. Verify the Docker context works:

   ```bash
   docker ps
   docker events
   ```

   Both commands work correctly.

3. Launch Lazydocker:

   ```bash
   lazydocker
   ```

4. Observe the error:

   ```text
   Docker event stream returned error:
   error during connect:
   Get "http://%2Ftmp%2Flazydocker-sshtunnel-XXXXXXXX/dockerhost.sock/v1.51/events":
   EOF
   ```

**Expected behaviour**

Lazydocker should connect to the remote Docker Engine and maintain the Docker events stream without errors, just as the Docker CLI does.

**Screenshots**

I can provide screenshots if needed. The error is displayed in the Lazydocker UI and references the `/v1.51/events` endpoint through Lazydocker's temporary SSH tunnel socket.

- OS: macOS Sequoia
- Lazydocker Version: v0.25.2
- Installed via Homebrew
- Not built from source

Docker Client:

```text
Version: 29.6.0
API version: 1.55
```

Remote Docker Server:

```text
Version: 27.5.1
API version: 1.47
```

**Additional context**

The issue appears specific to Lazydocker. The following all work correctly using the same Docker context:

```bash
docker ps
docker events
docker logs
docker version
```

The Docker context configuration is:

```text
Host: ssh://AORUS
```

where `AORUS` is an SSH host alias defined in `~/.ssh/config`.

I also tested connectivity directly and verified that Docker's SSH transport is functioning correctly.

The most interesting detail is that the error references:

```text
/v1.51/events
```

while the remote Docker daemon reports:

```text
API version: 1.47
```

Docker CLI successfully negotiates and communicates with the remote daemon, but Lazydocker appears to lose the event stream connection with an EOF error.

Please let me know if there is a debug mode or additional logging I can provide.

Also note these errors happen on both MacOS and Ubuntu Server.

hamodywe:
The Docker events stream is long-lived and can sit idle for a long time between events. Without SSH keepalives, an idle tunneled connection can be silently dropped by an intermediate NAT/firewall, or go undetected as dead by ssh itself, surfacing as an EOF on the tunneled socket once the events stream (or any other request) tries to use it. Short-lived requests like listing containers rarely hit this window, which is why only the events stream appeared broken.

Add -o ServerAliveInterval=15 -o ServerAliveCountMax=3 to the ssh tunnel command so idle-but-live connections stay open and dead ones are detected and torn down promptly.

Fixes #815


hamodywe:
﻿Closing this to keep my open-PR list honest - no hard feelings, I know maintainer time is the scarce thing. The branch stays on my fork, so if this is ever useful just say the word and I'll reopen it.

PR #834, opened 2026-08-23
Support multiple projects with the same service name (#883)

### LINKED ISSUE #833: Multiple projects with the same service names does not work
**Describe the bug**
Lazydocker does not support "clones" of the same docker compose project, or any case where two projects share a service name.

**To Reproduce**
Make `docker1` and `docker2` directories, each with this `compose.yaml`:
```
services:
  sleep:
    image: ubuntu:latest
    entrypoint: ["/bin/sh", "-c"]
    command: ["echo 'Sleeping...' && sleep 100000"]
    stop_grace_period: 0s
```

Run `docker compose create` in each. Now run lazydocker in each. One of them will detect the other as the project and will not be able to manage the one in the current directory. Using an explicit -p does not fix this.

**Expected behaviour**
Both instances manage their own local project regardless of the service name.

 - OS: Ubuntu
 - Lazydocker Version 0.25.2

**Additional context**
This was seemingly introduced in https://github.com/jesseduffield/lazydocker/pull/774

dougclayton:
Fixes https://github.com/jesseduffield/lazydocker/issues/833


dougclayton:
Full disclosure, this was done with the help of Copilot, but it's working for me locally.

